PR #37, opened 2022-12-07
Update to make CLI backwards compatible with Kube versions < 1.25

dsessler7:
[sc-16942]

### DIFF
--- a/internal/cmd/export.go
+++ b/internal/cmd/export.go
@@ -27,8 +27,10 @@ import (
 	"github.com/spf13/cobra"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
+	batchv1beta1 "k8s.io/api/batch/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -88,6 +90,17 @@ var namespacedResources = []schema.GroupVersionResource{{
 }, {
 	Version:  "v1",
 	Resource: "serviceaccounts",
+}}
+
+// These "removed" GVRs are for making our CLI backwards compatible with older PGO versions.
+var removedNamespacedResources = []schema.GroupVersionResource{{
+	Group:    batchv1beta1.SchemeGroupVersion.Group,
+	Version:  batchv1beta1.SchemeGroupVersion.Version,
+	Resource: "cronjobs",
+}, {
+	Group:    policyv1beta1.SchemeGroupVersion.Group,
+	Version:  policyv1beta1.SchemeGroupVersion.Version,
+	Resource: "poddisruptionbudgets",
 }}
 
 // newSupportCommand returns the support subcommand of the PGO plugin.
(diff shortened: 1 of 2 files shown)


internal/cmd/export.go
```diff
@@ -90,6 +92,17 @@ var namespacedResources = []schema.GroupVersionResource{{
 	Resource: "serviceaccounts",
 }}
 
+// These "removed" GVRs are for making our CLI backwards compatible with older Kube versions.
```
tjmoore4, line 95: :pick: If I understand correctly, I think it might be more accurate to say:
```suggestion
// These "removed" GVRs are for making our CLI backwards compatible with older PGO versions.
```
Kubernetes 1.25 support is what required PGO to use `v1` versions of CronJobs and PodDisruptionBudgets instead of the `v1beta1` versions, per
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#cronjob-v125
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125

testing/kuttl/e2e/version/00--check-version.yaml
```diff
@@ -2,22 +2,27 @@ apiVersion: kuttl.dev/v1beta1
...
+    VERSION_OUTPUT=$(kubectl pgo version)
 
     # the operator version isn't empty and the version output matches the CRD value
     if [[ -z $OPERATOR_VERSION || $VERSION_OUTPUT != *"Operator Version: v$OPERATOR_VERSION"* ]]; then
+      echo "Version output is: "
+      echo "$VERSION_OUTPUT"
+      echo "Expected: v$OPERATOR_VERSION"
       exit 1
-    fi
+    fi
```
tjmoore4, line 28: :pick: missing a trailing newline

tjmoore4 (approved):
Looks good to me, a couple of nitpicks.

testing/kuttl/e2e/version/00--check-version.yaml
```diff
@@ -2,22 +2,27 @@ apiVersion: kuttl.dev/v1beta1
 kind: TestStep
```
tjmoore4, line 2: :pick: One other thing I just thought about. It might be worth splitting this PR into two commits, one for the export specific stuff and one for the version test update. Not a blocker though.

internal/cmd/export.go
dsessler7: Yep, that's definitely what I meant, I just "misspoke". Good catch.

testing/kuttl/e2e/version/00--check-version.yaml
dsessler7: I need to leave a post-it note on my monitor: "Always leaving a trailing newline in YAML files"

testing/kuttl/e2e/version/00--check-version.yaml
dsessler7: I'll wait and see what @cbandy thinks about this.

testing/kuttl/e2e/version/00--check-version.yaml
cbandy: There's an "insert final newline" setting in VS Code. 🌱 We could explore editorconfig.org to commit preferences like this to our repos.

testing/kuttl/e2e/version/00--check-version.yaml
cbandy: ⛏️ This is a good case for a separate commit. Not a blocker here.

🌱 It's easier to squash "too many" commits than to ... split "too few." My recommendation is to tend toward "too many," even with terrible commit messages. They can be tidied or rearranged during review.

cbandy (approved):

PR #39, opened 2022-12-15
Update Kuttl tests to better handle variable error output

tjmoore4:
Update Kuttl tests to better handle variable error output.

### DIFF
--- a/testing/kuttl/e2e/backup/11--backup-with-new-flag.yaml
+++ b/testing/kuttl/e2e/backup/11--backup-with-new-flag.yaml
@@ -12,10 +12,7 @@ commands:
       exit 1
     fi
 
-    # FIXME(cbandy): why does the output start with a blank line?
-    ERROR=$(echo "${RESULT}" | head -n2)
-
-    if [[ "${ERROR,,}" != *conflict* || "${ERROR}" != *repoName* ]]; then
-      printf 'Expected conflict on repoName, got %q' "${ERROR}"
+    if [[ "${RESULT,,}" != *conflict* || "${RESULT}" != *repoName* ]]; then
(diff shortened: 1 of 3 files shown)


benjaminjb (approved):
GCP auth sometimes adds its own statements to returns/print statements, so this looks like a fair way to check returns with some slack for that

testing/kuttl/e2e/backup/11--backup-with-new-flag.yaml
```diff
@@ -12,10 +12,7 @@ commands:
       exit 1
     fi
 
-    # FIXME(cbandy): why does the output start with a blank line?
-    ERROR=$(echo "${RESULT}" | head -n2)
```
cbandy, line 16: 🤔 It seems like I was trying to get just the first line, but I don't recall why. Is there something else that normally prints on later lines?

cbandy (approved):
I would like it if our tests could be more focused/specific about the errors _we_ return, but I suppose with client-go doing whatever it wants to stderr we're kinda stuck.

testing/kuttl/e2e/backup/11--backup-with-new-flag.yaml
tjmoore4: From my tests, it looks like it normally prints the error this way:
```
Error requesting update: Apply failed with 1 conflict: conflict with "kubectl-kuttl" using postgres-operator.crunchydata.com/v1beta1: .spec.backups.pgbackrest.manual.repoName
Error: Apply failed with 1 conflict: conflict with "kubectl-kuttl" using postgres-operator.crunchydata.com/v1beta1: .spec.backups.pgbackrest.manual.repoName
```
The underlying error is the same, but the first time starts with `Error requesting update` and the second line starts with `Error`.

PR #40, opened 2022-12-20
Small adjustments to doc generation

cbandy:
I noticed that the Hugo alias on the root command directs away from "latest" at e.g. https://access.crunchydata.com/documentation/postgres-operator-client/latest/reference/pgo_backup/

At the same time, I moved the Go file into the `hack` directory so that `cmd` has only `kubectl-pgo`.


benjaminjb (approved):
I've answered my questions by just getting some time to focus and read

PR #42, opened 2022-12-28
add import for OpenID Connect auth

rnwgnr:
At the moment the pgo client will not work with k8s clusters that use OpenID Connect for authentication:
```
$ kubectl pgo version
Client Version: v0.2.0
Error: no Auth Provider found for name "oidc"
```

This PR adds the required module.

Tested with linux/amd64 so far. Unfortunately i don't have machines to test the remaining target platforms.


ValClarkson (approved):
Thanks for submitting!

PR #46, opened 2023-03-01
Add current timestamp to files in support export .tar file

tjmoore4:
Adds a timestamp to each file created for the support export .tar file. Value includes timezone information.

Issue: [sc-17845]


tjmoore4:
> Works and fixes the problem we were seeing (with files from the start of time).
> 
> One possibility that is probably unnecessary, but just want to bring it up: what if we saved the `time.Now()` result and used that in the tarball file name and in the tar headers so that they could be seen as matching? Meh, that's unnecessary.

I thought about that but, on further consideration, I decided it made more sense to preserve as much detail for the timestamp as possible. There is a chance (albeit small) that a change happens in the seconds it takes for the command to run and it would therefore be helpful to know exactly when the information was gathered. Plus, this way is also simpler, which doesn't hurt...

PR #47, opened 2023-03-06
Update Node list file output and include Node spec info

tjmoore4:
Updates the generation of the Node 'list' file to include comparable information to the 'kubectl get node -o wide' command. Also adds the yaml output of each Node in the cluster. Finally, the 'support- export' test was updated to verify the correct behavior.

Issue: [sc-17846]


dsessler7 (approved):
LGTM

PR #48, opened 2023-03-08
Add monitoring Pod log capture and improve associated tests

tjmoore4:
This update adds the logs from the Grafana, Prometheus and Alertmanager Pods to the support export archive. By default, the support export command will look in the same namespace as the given PostgresCluster. For cases where the monitoring Pods run in a separate namespace, an override flag, --monitoring-namespace, is provided. Note that monitoring logs are only gathered from a single namespace at a time.

Additionally, this update adds associated tests for the monitoring logs to include the Pods mentioned above and the Postgres exporter sidecar container when defined in the PostgresCluster spec.

Issues:
[sc-17848]
[sc-17691]

### DIFF
--- a/docs/content/reference/pgo_support_export.md
+++ b/docs/content/reference/pgo_support_export.md
@@ -46,14 +46,19 @@ pgo support export CLUSTER_NAME [flags]
   
   # Long Flags
   kubectl pgo support export daisy --output . --pg-logs-count 2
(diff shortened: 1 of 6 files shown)


tony-landreth (approved):

dsessler7 (approved):
LGTM

PR #49, opened 2023-03-09
Override Prometheus container command

tjmoore4:
This change overrides the Prometheus container's default command to avoid a 'permission denied' error in some environments (e.g. Openshift).

This happens because the Kuttl test is defining a 'dummy' deployment that is not configured for actual use. As such, overriding the command does not impact the validity of the test.

Issue: [sc-17848]

### DIFF
--- a/testing/kuttl/e2e/support-export/20--create_cluster_with_monitoring.yaml
+++ b/testing/kuttl/e2e/support-export/20--create_cluster_with_monitoring.yaml
@@ -38,6 +38,9 @@ spec:
       containers:
       - image: prom/prometheus:v2.33.5
         name: prometheus
+        # Override default command to avoid 'permission denied' error in some
+        # environments (e.g. Openshift). Echo 'hello' so the Pod log is not empty.
+        command: ["sh", "-c", "while true; do echo hello; sleep 10;done"]
 ---
 apiVersion: apps/v1
 kind: Deployment


ValClarkson (approved):

testing/kuttl/e2e/support-export/20--create_cluster_with_monitoring.yaml
```diff
@@ -38,6 +38,10 @@ spec:
       containers:
       - image: prom/prometheus:v2.33.5
         name: prometheus
+        # override default command to avoid 'permission denied' error in some
+        # environments (e.g. Openshift)
+        command: ["/bin/sh"]
+        args: ["-c", "while true; do echo hello; sleep 10;done"]
```
tony-landreth, line 44: Why do we echo hello?

tjmoore4: So that we'll have something in the logs. It doesn't matter _what's_ there, but we want something to be there.

cbandy, line 44: ⛏️ `sh` might be more future-proof.

🤔 I forget; does setting `command` implicitly override the image args? If so, these could be combined into one array.

benjaminjb: Might want to add that to the comment?

tjmoore4: :+1: Comment is updated.

benjaminjb (approved):

PR #50, opened 2023-03-10
Add E2E Tests to PR Testing

tony-landreth:
This commit adds a workflow for running our kuttl tests in a GitHub action.

Issue: [sc-17724]

### DIFF
--- a/.github/actions/k3d/action.yaml
+++ b/.github/actions/k3d/action.yaml
@@ -1,0 +1,31 @@
+name: k3d
+description: Start k3s using k3d
+inputs:
+  k3d-tag:
+    default: latest
+    required: true
+    description: >
+      Git tag from https://github.com/k3d-io/k3d/releases or "latest"
+  k3s-channel:
+    default: latest
+    required: true
+    description: >
+      https://docs.k3s.io/upgrades/manual#release-channels
+
+runs:
+  using: composite
+  steps:
+    - id: k3d
+      name: Install k3d
+      shell: bash
+      env:
+        K3D_TAG: ${{ inputs.k3d-tag }}
+      run: |
+        curl --fail --silent https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh |
+          TAG="${K3D_TAG#latest}" bash
+
+    - id: k3s
+      name: Start k3s
+      shell: bash
+      run: |
+        k3d cluster create --image '+${{ inputs.k3s-channel }}' --no-lb --timeout=2m --wait
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -18,3 +18,43 @@ jobs:
 
       - run: make check
       - run: make check-cli-docs
+
+  kuttl-test:
+    runs-on: ubuntu-latest
+    needs: [go-test]
+    strategy:
+      fail-fast: false
+      matrix:
+        kubernetes: [v1.25, v1.24, v1.23, v1.22, v1.21]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with: { go-version: 1.x }
+
+      - name: Start k3s
+        uses: ./.github/actions/k3d
+        with:
+          k3s-channel: "${{ matrix.kubernetes }}"
+
+      - name: Build CLI
+        run: make build
+
+      - name: Create operator namespace
+        run: kubectl create namespace postgres-operator
+
+      - name: Install Operator
+        run: helm install pgo oci://registry.developers.crunchydata.com/crunchydata/pgo -n postgres-operator
+
+      - name: Use bash, so we can use double square-bracket [[ syntax in our scripts
+        run: |
+          sudo ln -s bash /bin/sh.bash
+          sudo mv /bin/sh.bash /bin/sh
+
+      - name: Install kuttl
+        run: |
+          curl -Lo /usr/local/bin/kubectl-kuttl https://github.com/kudobuilder/kuttl/releases/download/v0.15.0/kubectl-kuttl_0.15.0_linux_x86_64
+          chmod +x /usr/local/bin/kubectl-kuttl
+
+      - name: Run kuttl
+        run: |
+          make check-kuttl


.github/workflows/test.yaml
```diff
@@ -18,3 +18,41 @@ jobs:
...
       - run: make check
       - run: make check-cli-docs
+
+  kuttl-test:
+    runs-on: ubuntu-latest
+    needs: [go-test]
+    strategy:
+      fail-fast: false
+      matrix:
+        kubernetes: [v1.25, v1.24, v1.23, v1.22, v1.21]
```
tony-landreth, line 28: I'd be inclined to trim this list down a little -- maybe 3 versions -- to see if it reduces the number of flakes.

benjaminjb: We have this issue in postgres-operator too, that our PR commits will fail tests for reasons of flakiness; I'm inclined to test more, even with flakes, and put getting the flakes out higher on our priority list. (It's almost too bad that GH makes rerunning failed tests so easy...)

Unless what you're suspecting is that our flaking tests is a result of tests ... jamming each other?

.github/workflows/test.yaml
```diff
@@ -18,3 +18,41 @@ jobs:
...
+      - name: Create operator namespace
+        run: kubectl create namespace postgres-operator
+
+      - name: Install Operator
+        run: helm install pgo oci://registry.developers.crunchydata.com/crunchydata/pgo -n postgres-operator
+
+      - name: Use bash
+        run: |
+          sudo ln -s bash /bin/sh.bash
+          sudo mv /bin/sh.bash /bin/sh
```
benjaminjb, line 49: 🔧 Can you comment the reason behind this?

.github/workflows/test.yaml
tony-landreth: I like testing more, even with flakes. I'll leave it.

.github/workflows/test.yaml
```diff
@@ -18,3 +18,41 @@ jobs:
 
       - run: make check
       - run: make check-cli-docs
+
+  kuttl-test:
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
tony-landreth: I think it might be good to let it drift. If the latest Ubuntu breaks us, that might be informative.

.github/workflows/test.yaml
```diff
@@ -18,3 +18,41 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        kubernetes: [v1.25, v1.24, v1.23, v1.22, v1.21]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with: { go-version: 1.x }
+
+      - name: Start k3s
+        uses: ./.github/actions/k3d
```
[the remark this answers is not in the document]
tony-landreth: Good point.

.github/workflows/test.yaml
```diff
@@ -18,3 +18,41 @@ jobs:
...
+
+      - name: Build CLI
+        run: make build
+
+      - name: Create operator namespace
+        run: kubectl create namespace postgres-operator
+
+      - name: Install Operator
+        run: helm install pgo oci://registry.developers.crunchydata.com/crunchydata/pgo -n postgres-operator
```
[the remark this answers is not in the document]
tony-landreth: We have a plan to test other versions elsewhere.

.github/workflows/test.yaml
tony-landreth: Good call. I'll add a comment. I'll also mention it here: Ubuntu defaults to dash, rather than bash, which doesn't recognize our use of `[[ ]]` syntax.

.github/workflows/test.yaml
```diff
@@ -18,3 +18,41 @@ jobs:
...
+
+      - name: Use bash
+        run: |
+          sudo ln -s bash /bin/sh.bash
+          sudo mv /bin/sh.bash /bin/sh
+
+      - name: Install kuttl
+        run: |
+          curl -Lo /usr/local/bin/kubectl-kuttl https://github.com/kudobuilder/kuttl/releases/download/v0.13.0/kubectl-kuttl_0.13.0_linux_x86_64
```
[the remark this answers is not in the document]
tony-landreth: I looked at installing krew to grab the latest kuttl, but it proved a little tricky.

.github/actions/k3d/action.yaml
```diff
@@ -0,0 +1,40 @@
...
+      run: |
+        k3d cluster create --image '+${{ inputs.k3s-channel }}' --no-lb --timeout=2m --wait
+        kubectl version --short | awk '{ print "${tolower($1)}=${$3}" >> $GITHUB_OUTPUT }'
+
+        PAUSE_IMAGE=$(docker exec $(k3d node list --output json | jq --raw-output 'first.name') \
+          k3s agent --help | awk '$1 == "--pause-image" {
+            match($0, /default: "[^"]*"/);
+            print substr($0, RSTART+10, RLENGTH-11)
+          }')
+        echo "pause-image=${PAUSE_IMAGE}" >> $GITHUB_OUTPUT
```
benjaminjb, line 40: 🤔  Not using this right now, which is fine with me: if we don't need to prefetch the pause-image to pass the tests, then... hmmm, well, ok, let's drop it for now and see how it goes?

benjaminjb (approved):
LGTM, just one thought re: an output that we don't seem to be using, tests look good.

.github/actions/k3d/action.yaml
tony-landreth: I meant to drop this one. Thanks for catching it!

.github/actions/k3d/action.yaml
```diff
@@ -0,0 +1,32 @@
...
+  steps:
+    - id: k3d
+      name: Install k3d
+      shell: bash
+      env:
+        K3D_TAG: ${{ inputs.k3d-tag }}
+      run: |
+        curl --fail --silent https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh |
+          TAG="${K3D_TAG#latest}" bash
+        k3d version | awk '{ print "${tolower($1)}=${$3}" >> $GITHUB_OUTPUT }'
```
benjaminjb, line 26: Are we using this for anything now?

tony-landreth: I don't see how we could be. I'll delete that.

benjaminjb (approved):

PR #51, opened 2023-03-13
Accurate event recording by support export

tjmoore4:
This update corrects the current behavior of the support export command's event capture, adds relevant Kuttl checks and documents Kubernetes Event storage duration.

Issue: [sc-17993]

### DIFF
--- a/docs/content/reference/pgo_support_export.md
+++ b/docs/content/reference/pgo_support_export.md
@@ -34,6 +34,12 @@ PostgresCluster.
 
     Note: This RBAC needs to be cluster-scoped to retrieve information on nodes.
 
+#### Event Capture
+    Support export captures all Events in the PostgresCluster's Namespace.
+    Event duration is determined by the '--event-ttl' setting of the Kubernetes
+    API server. Default is 1 hour.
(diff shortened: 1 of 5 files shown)


docs/content/reference/pgo_support_export.md
```diff
@@ -33,6 +33,10 @@ PostgresCluster.
     statefulsets.apps                                   [get list]
 
     Note: This RBAC needs to be cluster-scoped to retrieve information on nodes.
+	
+#### Event Capture
+	Event duration is determined by the '--event-ttl' setting of the Kubernetes API server. Default is 1 hour.
```
benjaminjb, line 38: (a) 👍 
(b) I want to make a change to `internal/cmd/pgo.go` so this gets formatted correctly in our docs and in our `--help` output -- I can push up that change or meet up to discuss
(c) I think we should note that events aren't limited to the cluster, but will grab all events in the ns, yeah?

tjmoore4: :+1: Updated per discussion.

benjaminjb (approved):

PR #53, opened 2023-03-14
Support Export Compression

tjmoore4:
This update changes the support export command to produce a compressed .tar.gz file instead of an uncompressed .tar file. Relevant tests are updated accordingly.

Issue: [sc-18010]

### DIFF
--- a/internal/cmd/export.go
+++ b/internal/cmd/export.go
@@ -17,6 +17,7 @@ package cmd
 import (
 	"archive/tar"
 	"bytes"
+	"compress/gzip"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


internal/cmd/export.go
```diff
@@ -226,18 +227,25 @@ kubectl pgo support export daisy --monitoring-namespace another-namespace --outp
...
+		// Example: crunchy_k8s_support_export_2022-08-08-115726-0400.tar.gz
+		outputFile := "crunchy_k8s_support_export_" + time.Now().Format("2006-01-02-150405-0700") + ".tar.gz"
 		// #nosec G304 -- We intentionally write to the directory supplied by the user.
 		tarFile, err := os.Create(outputDir + "/" + outputFile)
 		if err != nil {
 			return err
 		}
 
-		tw := tar.NewWriter(tarFile)
+		gw, err := gzip.NewWriterLevel(tarFile, gzip.BestCompression)
```
jmckulk, line 238: Does the order that you create these writers make a difference? Why did you create the GZIP writer first, then the tar writer?

I'm not familiar with what happens when you layer writers like this.

testing/kuttl/e2e/support-export/01--support_export.yaml
```diff
@@ -3,14 +3,23 @@ apiVersion: kuttl.dev/v1beta1
...
 
     DIR="./kuttl-support-cluster/nodes/"
     LIST="${DIR}list"
 
-    CLEANUP="rm -r ./kuttl-support-cluster ./crunchy_k8s_support_export_*.tar"
+    CLEANUP="rm -r ./kuttl-support-cluster ./crunchy_k8s_support_export_*.tar.gz"
+
+    # check for expected gzip compression level
+    FILE_INFO=$(file ./crunchy_k8s_support_export_*.tar.gz)
+    [[ "${FILE_INFO}" == *'gzip compressed data, max compression'* ]] || {
```
jmckulk, line 17: I like the test :+1:

testing/kuttl/e2e/support-export/01--support_export.yaml
jmckulk, line 1: Would it make sense to rename this test now that you are updating/adding other tests?

Maybe something that references archiving or compression testing? :pick:

jmckulk (approved):
LGTM

internal/cmd/export.go
tjmoore4: Yes, the order matters because we want to end up with a "Tar archive (gzip-compressed)" file, not the other way around. (Technically, you can swap the order since both writers take an `io.Writer`, but then the `writeTar` function would need to be rewritten (at least) and I'm not positive what the final result would be.)

testing/kuttl/e2e/support-export/01--support_export.yaml
tjmoore4: :thinking: I'm not sure it should be renamed at this point. This is still the 'normal' use case test since there will no longer be a non-compressed version. At some point it may make sense to break it up, but for now I'm not sure it's necessary.

PR #54, opened 2023-03-15
Support export size detail message / PGO CLI version in export

tjmoore4:
With this update, the support export command now reports the resulting archive file's size along with usage instructions. In cases where the file is larger than 25MiB, an additional note is displayed with further instructions due to common email attachment size limits.

A second commit adds the PGO CLI version to the information collected by the support export command.
    
Issue: [sc-18005]
Issue: [sc-18016]


dsessler7 (approved):
LGTM

PR #55, opened 2023-03-17
Support export limitrange ingress

tjmoore4:
This PR does the following:

- Removes the client 'get' API calls from the gatherNamespacedAPIResources function and instead uses the object returned to the 'list' object previously made. Associated CLI help output and documentation are also updated.

- Adds the collection of any LimitRanges or Ingresses defined in the PostgresCluster's Namespace to the support export command archive output.

- Update the clusterNamespacedResources variable's Group and Version values to a consistent definition for all Resources.
    
Issues:
[sc-18056]
[sc-18061]


testing/kuttl/e2e/support-export/22--cleanup.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: kuttl.dev/v1beta1
```
benjaminjb, line 1: 🤔 Might it be a faster test if we (1) skip deleting the cluster, (2) add the limitrange and ingress, and (3) run the command and see if our files are there?

(Also 🤔 , do we want to alter the original test so that it makes sure we don't have any limitrange or ingress?)

docs/content/reference/pgo_support_export.md
```diff
@@ -13,24 +13,26 @@ PostgresCluster.
 #### RBAC Requirements
     Resources                                           Verbs
     ---------                                           -----
-    configmaps                                          [get list]
-    cronjobs.batch                                      [get list]
-    deployments.apps                                    [get list]
-    endpoints                                           [get list]
+    configmaps                                          [list]
```
benjaminjb, line 16: Have you run this with `--as system:serviceaccount:<namespace>:<serviceaccount>` with an rbac scoped to this just to make sure it works?

testing/kuttl/e2e/support-export/31--support_export.yaml
```diff
@@ -0,0 +1,56 @@
...
+- script: tar -xf ./crunchy_k8s_support_export_*.tar.gz
+- script: |
+    #! /bin/bash
+
+    # LimitRange directory and list file path
+    LR_DIR="./kuttl-support-cluster-other/limitranges/"
+    LR_LIST="${LR_DIR}list"
+
+    # check for limitrange object name
+    LRO=$(awk 'NR==2 {print $1}' "${LR_LIST}")
```
benjaminjb, line 15: If this path doesn't exist, this will blow up -- what happens in that case?

benjaminjb (commented):
Overall looks good, a couple of questions about the test -- and I think a readme in the test would help at the current size.

testing/kuttl/e2e/support-export/31--support_export.yaml
```diff
@@ -0,0 +1,56 @@
...
+    [[ "${LRO}" == 'storagelimits' ]] || {
+      echo "Expected 'storagelimits' limitrange, got:"
+      echo "${LRO}"
+      eval "$CLEANUP"
```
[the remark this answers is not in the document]
tjmoore4: Nope, that was an oversight during refactor. Good catch; I'll correct this.

docs/content/reference/pgo_support_export.md
tjmoore4: In a slightly different way, yes. I used the `--kubeconfig` flag to pass in a configuration that ultimately pointed to a custom ClusterRole. In testing in a couple different environments, everything worked as expected. Happy to talk more about this in case there are other testing options you think would be useful.

testing/kuttl/e2e/support-export/31--support_export.yaml
tjmoore4: The test fails with an `exit 1`, which is good, but I do see that this results in the cleanup code not being executed, which isn't great. I'll add some "file exists" checks to handle this case better.

testing/kuttl/e2e/support-export/22--cleanup.yaml
tjmoore4: :thinking: I could see that. I think it would make the most sense to refactor so that the 'monitoring' cluster is last though and use the original cluster for the first and second tests (because the monitoring cluster is the more unique one). I'll work on an implementation that makes sense here.

tjmoore4: I pushed the test refactor along with a README overview. I'm not thinking we should remove any existing LimitRanges or Ingresses in case the tests are run in existing Namespaces, but I did rename the LimitRange and Ingress so they're more obviously related to the test run.

docs/content/reference/pgo_support_export.md
benjaminjb: Oh, no need to run it that particular way, that's just the way I found easy to test when doing the original RBAC story, and just wanted to make sure it was tested.

testing/kuttl/e2e/support-export/21--support_export.yaml
```diff
@@ -2,42 +2,57 @@
 apiVersion: kuttl.dev/v1beta1
 kind: TestStep
 commands:
-- script: kubectl-pgo --namespace $NAMESPACE support export kuttl-support-monitoring-cluster -o .
+- script: kubectl-pgo --namespace $NAMESPACE support export kuttl-support-cluster -o .
 - script: tar -xf ./crunchy_k8s_support_export_*.tar.gz
 - script: |
-    #!/bin/bash
+    #! /bin/bash
```
benjaminjb, line 8: 🔧 ?

testing/kuttl/e2e/support-export/README
```diff
@@ -0,0 +1,78 @@
...
+#### Basic Cluster
+
+- 00--create_cluster.yaml
+    Create PostgresCluster `kuttl-support-cluster` using the PGO CLI.
+
+- 00-assert.yaml
+    Assert `kuttl-support-cluster` deploys successfully.
+
+- 01--support_export.yaml
+    Create and verifies the support export archive. Current checks are made for
```
benjaminjb, line 19:
```suggestion
    Create and verify the support export archive. Current checks are made for
```
?

testing/kuttl/e2e/support-export/README
```diff
@@ -0,0 +1,78 @@
...
+    Create PostgresCluster `kuttl-support-monitoring-cluster` with Exporter
+    sidecar and a dummy Monitoring stack, i.e. fake `crunchy-prometheus`,
+    `crunchy-grafana` and `crunchy-alertmanager` Deployments.
+
+- 30-assert.yaml
+    Assert `kuttl-support-monitoring-cluster` and fake monitoring Deployments
+    deploy successfully.
+
+- 31--support_export.yaml
+    Verify log files are created for the 4 monitoring related Containers, i.e.
```
benjaminjb, line 67:
```suggestion
    Create and verify support export log files are created for the 4 monitoring related Containers, i.e.
```?

File: testing/kuttl/e2e/support-export/README [hunk 8]
```diff
@@ -0,0 +1,78 @@
...
+#### LimitRange and Ingress
+
+- 20--create_limitrange_ingress.yaml
+    Create a LimitRange and Ingress
+
+- 20-assert.yaml
+    Assert LimitRange and Ingress exist
+
+- 21--support_export.yaml
+    Create and verifies support export archive. Current checks are made for
```
benjaminjb, line 41:
```suggestion
    Create and verify support export archive. Current checks are made for
```

benjaminjb (approved):
Really like that README layout, I might want to update some of our other readmes accordingly (later)

testing/kuttl/e2e/support-export/README
tjmoore4: Updated to
```
    Create and verify the support export archive. Verify log files are created
    for the 4 monitoring related Containers, i.e.
```

testing/kuttl/e2e/support-export/README
tjmoore4: :+1: updated.

benjaminjb (approved):

PR #56, opened 2023-03-22
Capture process information in support export archive

tjmoore4:
This update adds the capture of running processes for each Pod of the selected PostgresCluster. The 'ps' command is run on any possible container in the cluster, but only successful results are stored in the archive.

Issue: [sc-18017]

### DIFF
--- a/internal/cmd/export.go
+++ b/internal/cmd/export.go
@@ -360,6 +360,11 @@ kubectl pgo support export daisy --monitoring-namespace another-namespace --outp
 			err = gatherPatroniInfo(ctx, clientset, restConfig, namespace, clusterName, tw, cmd)
 		}
 
+		// Exec to get Container processes
+		if err == nil {
+			err = gatherProcessInfo(ctx, clientset, restConfig, namespace, clusterName, tw, cmd)
+		}
+
 		// Print cli output
 		path := clusterName + "/logs/cli"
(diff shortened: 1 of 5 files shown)


internal/cmd/export.go
```diff
@@ -931,6 +936,80 @@ func gatherPatroniInfo(ctx context.Context,
...
+// running process information for each Pod.
+func gatherProcessInfo(ctx context.Context,
+	clientset *kubernetes.Clientset,
+	config *rest.Config,
+	namespace string,
+	clusterName string,
+	tw *tar.Writer,
+	cmd *cobra.Command,
+) error {
+	// TODO: update to use specific client after SSA change
```
tony-landreth, line 949: :question: What is SSA?

internal/cmd/export.go
```diff
@@ -931,6 +936,80 @@ func gatherPatroniInfo(ctx context.Context,
...
+	}
+
+	podExec, err := util.NewPodExecutor(config)
+	if err != nil {
+		return err
+	}
+
+	for _, pod := range pods.Items {
+		for _, container := range pod.Spec.Containers {
+			// Attempt to exec in and run 'ps' command in all available containers,
```
tony-landreth, line 975: Helpful comment :+1:

internal/cmd/export.go
tjmoore4: After further discussion with the team, seems this TODO is no longer necessary. I'm removing it (along with the other two existing TODOs).

PR #58, opened 2023-03-28
Improve support export self-logging

tjmoore4:
This update improves the support export self-logging with timestamp information in the log file, captured arguments and flags, current Kubernetes context name and streamlined output when the command is executed.

Issue: [sc-18066]

### DIFF
--- a/internal/cmd/export.go
+++ b/internal/cmd/export.go
@@ -56,14 +56,22 @@ const (
 	// - https://mathworld.wolfram.com/Mebibyte.html
 	mebibyte float64 = (1 << 20)
 
+	// formatting for CLI log and stdout
+	preBox  = "┌────────────────────────────────────────────────────────────────"
+	postBox = "└────────────────────────────────────────────────────────────────"
+
 	// Default support export message
-	msg1 = "\n" + `Archive file size: %.2f MiB
-Email the support export archive to support@crunchydata.com
-or attach as a email reply to your existing Support Ticket` + "\n"
+	msg1 = "\n" + `| Archive file size: %.2f MiB
+| Email the support export archive to support@crunchydata.com
+| or attach as a email reply to your existing Support Ticket` + "\n"
 
 	// Additional support export message. Shown when size is greater than 25 MiB.
-	msg2 = `Archive file (%.2f MiB) may be too big to email.
-Please request file share link by emailing support@crunchydata.com` + "\n"
+	msg2 = "\n" + `| Archive file (%.2f MiB) may be too big to email.
+| Please request file share link by emailing
+| support@crunchydata.com` + "\n"
+
+	// time formatting for CLI logs
+	logTimeFormat = "2006-01-02 15:04:05.000 -0700 MST"
 )
 
 // namespaced resources that have a cluster Label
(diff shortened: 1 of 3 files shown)


internal/cmd/export.go
```diff
@@ -218,7 +225,18 @@ kubectl pgo support export daisy --monitoring-namespace another-namespace --outp
...
 
+		writeInfo(cmd, preBox)
+		writeInfo(cmd, "| PGO CLI Support Export Tool")
+		writeInfo(cmd, "| The support export tool will collect information that is")
+		writeInfo(cmd, "| commonly necessary for troubleshooting a PostgresCluster.")
+		writeInfo(cmd, "| Note: No data or k8s secrets are collected.")
+		writeInfo(cmd, postBox)
+
 		clusterName := args[0]
+		cmd.Printf("%sArg - PostgresCluster Name: %s\n", debugPrefix(), clusterName)
```
jmckulk, line 236: I think I would appreciate a `writeDebug()` function to match `writeInfo` :thinking: :pick: 

It would also be more consistent with a logger if we implement one :thought_balloon:

internal/cmd/export.go
```diff
@@ -663,11 +715,12 @@ func gatherEvents(ctx context.Context,
 	tw *tar.Writer,
 	cmd *cobra.Command,
 ) error {
+	writeInfo(cmd, "Collecting events...")
 	// TODO (jmckulk): do we need to order events?
```
jmckulk, line 719: You have been looking at events more than I did at the time. Is this `TODO` still relevant? Maybe it could be removed

internal/cmd/export.go
```diff
@@ -183,10 +191,9 @@ PostgresCluster.
 	// Set output to log and write to buffer for writing to file
 	var cliOutput bytes.Buffer
 	cmd.PreRunE = func(cmd *cobra.Command, args []string) error {
-		outMW := io.MultiWriter(os.Stdout, &cliOutput)
 		errMW := io.MultiWriter(os.Stderr, &cliOutput)
-		cmd.SetOut(outMW)
 		cmd.SetErr(errMW)
+		cmd.SetOut(&cliOutput)
```
jmckulk, line 196: It would be helpful to have a comment describing why this change was made and where info and debug lines end up. I know we talked about it the other day, but I already forget the specifics.

jmckulk (approved):
This all LGTM. There are a couple of things that I would like to see changed, but nothing to block approval :+1:

internal/cmd/export.go
```diff
@@ -56,14 +56,22 @@ const (
...
+	msg1 = "\n" + `| Archive file size: %.2f MiB
+| Email the support export archive to support@crunchydata.com
+| or attach as a email reply to your existing Support Ticket` + "\n"
 
 	// Additional support export message. Shown when size is greater than 25 MiB.
-	msg2 = `Archive file (%.2f MiB) may be too big to email.
-Please request file share link by emailing support@crunchydata.com` + "\n"
+	msg2 = "\n" + `| Archive file (%.2f MiB) may be too big to email.
+| Please request file share link by emailing
+| support@crunchydata.com` + "\n"
```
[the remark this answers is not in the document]
tjmoore4: To my eye, it looks a little weird and stretched out that way. Admittedly, the way I have it has it's drawbacks, but it's fewer lines overall. If you're ok with it, I think I'd prefer to keep it as is.

internal/cmd/export.go
tjmoore4: :+1: updated.

internal/cmd/export.go
tjmoore4: The order looks like it matches what you get from `kubectl get events`, so I think it's safe to remove.

internal/cmd/export.go
tjmoore4: Comments added.

PR #61, opened 2023-04-13
Add the ability to shutdown/start PostgresClusters

boris-lenzinger:
**Why this pull request ?**
When using crunchy clusters, I had the need to stop some clusters (or all pg clusters) of a kubernetes cluster. To avoid manual updates, I needed to add the feature to the  pgo command line.

**What is added ?**
A subcommand `power` has been added with 2 subcommands: `on` and `off`. One can:
- stop/start a specific cluster is a given namespace
- stop/start a list of clusters in a given namespace
- stop/start all the clusters in a namespace
- stop/start all the clusters in a kubernetes cluster

Kuttl tests have been added to validate the process and some unit tests.


tjmoore4:
@boris-lenzinger Thank you for this contribution. The ability to shut down or start clusters would be a great feature to add. We would, however, like to keep the terminology around this process consistent with the spec. So instead of using `power on` or `power off` we would prefer something related to `shutdown` or `start`.  What do you think about the following commands?

```
pgo update <cluster> --shutdown=true/false
pgo update <cluster> --paused=true/false
```

I should note, we aren't sold on this idea yet, but it would allow us to update various pieces of the spec, like shutdown and paused, without having individual commands for each. Similarly, for the show command, the word `status` has a specific meaning in Kubernetes that includes more than just if a cluster is on or off.

Lastly, in general, more atomic changes will be easier for us to review and get merged. Updates like the `.idea` directory and the command colors should ideally be split out into a separate PR.

Once we decide on the implementation, would you be open to updating this PR with our feedback?

boris-lenzinger:
@tjmoore4 thank you for taking this into account.

Of course I'm open to make the changes when you have decided about the syntax. Not a problem at all.

About the syntax, I would prefer something more expressive and hide the values from the user. But it really depends on the way you want to promote this tool. If this is kind of "expert" oriented, definitely, values deserve to be exposed to the end-user. If you want to keep it as a utility tool for "normal" user, may be they will be confused about the data they can pass.

For boolean value, it feels a little bit cumbersome to pass the value on the command line but I see your point to easily update some fields of the spec.

Sorry for commiting .idea folder, this was a mistake ! I have another merge request that I would like to post about making a clone from a source cluster but it is quite a large one and may be we would need to talk about it before I submit it.

Thank you for your feedback. Keep me up-to-date of your decision about the syntax and I will make the changes the way you want them to be done.

benjaminjb:
@boris-lenzinger We appreciate your pull request and your interest in helping the CLI grow. Ultimately, we ended up going in a different direction in the implementation of the shutdown/start feature. That being said, we now have commands that are functionally similar and hope that these new features meet your needs. Thanks.

If you have other suggestions for the CLI or other feature requests -- or want to contribute code but want to talk over the idea first -- please don't hesitate to reach out to us!

PR #72, opened 2023-10-24
update actions v3->v4

benjaminjb:


benjaminjb:
nvm, the v4 of these actions aren't set to use Node.js 20.

PR #73, opened 2023-10-25
Require a Postgres version when creating a PostgresCluster

tjmoore4:
This change adds a new flag 'pg-major-version' to the 'create postgrescluster' command to allow a specific Postgres version to
be used for the PostgresCluster. The new flag is required and does not default to a particular version. As before, the image used by the cluster will be pulled from the corresponding 'RELATED_IMAGE' value set in the environment.

KUTTL test updates are also included, along with a bump to Postgres 15 and a timeout adjustment.

Finally, as a separate commit, this PR improves error printing behavior and ensure a backup annotation race
condition won't happen in faster test environments.

Issue: PGO-446


internal/cmd/create.go
```diff
@@ -59,9 +60,13 @@ func newCreateClusterCommand(config *internal.Config) *cobra.Command {
 
 	cmd.Args = cobra.ExactArgs(1)
 
+	var pgMajorVersion int
+	cmd.Flags().IntVarP(&pgMajorVersion, "pg-major-version", "", 0, "Set the Postgres major version")
```
benjaminjb, line 64: If you use `IntVar`, you don't need to supply a shorthand.

testing/kuttl/e2e/create/01--check_version_validation.yaml
```diff
@@ -0,0 +1,46 @@
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
```
benjaminjb, line 5: I think you could test these not through KUTTL: https://gianarb.it/blog/golang-mockmania-cli-command-with-cobra

benjaminjb: Wait, no, these errors are being returned by PGO -- the cli isn't generating anything in most of these cases, just passing through... how much "passing through" should we test?

benjaminjb: hmmm, maybe drop the too-low/too-high tests -- that's just what we get from PGO -- and move the other tests into the unit-testing form?

(or could leave the too-low/too-high tests here: to me, we do a lot of relying on PGO in the pgo-client, and that's OK. I don't mind a belt-and-suspenders approach, where we trust but verify.)

tjmoore4: I did some testing it looks like testing this behavior elsewhere would take a good bit of effort and possibly some refactoring. I think it still may be worth dropping the 'pass-through' tests, but otherwise I think KUTTL makes the most sense for now. Happy to discuss further though if there are other options.

benjaminjb (approved):

testing/kuttl/e2e/create/01--check_version_validation.yaml
```diff
@@ -0,0 +1,46 @@
...
+     # Verify the error when the Postgres version given is too low.
```
tony-landreth, line 6: Since too low is just > 10, A user can still "create" a cluster with PG 12, even if they don't have a PG 12 available. No pods will be brought up, but for example:
```
> ./bin/kubectl-pgo --namespace=postgres-operator create postgrescluster elephant --pg-major-version 12
postgresclusters/elephant created
> ./bin/kubectl-pgo --namespace=postgres-operator create postgrescluster elephant --pg-major-version 13
Error: postgresclusters.postgres-operator.crunchydata.com "elephant" already exists
```
It seems intuitive what to do here. Just thought I'd mention it.

tjmoore4: :+1: Yeah, with the existing event warning we have in place, it should be fairly clear what's going on, but we can definitely try to improve this in the future, especially if we make other updates.

tony-landreth (approved):

PR #75, opened 2023-10-30
PostgresCluster names and namespaces for support

tony-landreth:
Adds all reachable postgrescluster names and their namespaces to support export.

Issue: PGO-272


docs/content/reference/pgo_support_export.md
```diff
@@ -71,6 +71,7 @@ kubectl pgo support export daisy --monitoring-namespace another-namespace --outp
 | Note: No data or k8s secrets are collected.
 └────────────────────────────────────────────────────────────────
 Collecting PGO CLI version...
+Collecting names and namespaces for PostgresClusters...
```
benjaminjb, line 74: 💯

internal/cmd/export.go
```diff
@@ -484,6 +501,38 @@ func gatherPGOCLIVersion(_ context.Context,
 	return nil
 }
 
+func gatherPostgresClusterNames(clusterName string, ctx context.Context, cmd *cobra.Command, tw *tar.Writer, clientset *kubernetes.Clientset) error {
```
benjaminjb, line 504: would this work with the PostgresClient (instantiated ~327)?

tony-landreth: It does!

benjaminjb (approved):

internal/cmd/export.go
```diff
@@ -484,6 +490,40 @@ func gatherPGOCLIVersion(_ context.Context,
...
+		if err != nil {
+			return err
+		}
+		name, found, err := unstructured.NestedString(item.Object, "metadata", "name")
+		if !found {
+			return fmt.Errorf("key not found: metadata.name")
+		}
+		if err != nil {
+			return err
+		}
```
cbandy, line 515: 🌱 Every PostgresCluster returned from the API will have a namespace and name. There's also no harm in proceeding if we _do_ encounter something unexpected. I think we prefer to gather the remainder of the list rather than give up early.

```suggestion
		ns, _, _ := unstructured.NestedString(item.Object, "metadata", "namespace")
		name, _, _ := unstructured.NestedString(item.Object, "metadata", "name")
```

PR #76, opened 2023-10-30
Roughest draft of Create separation + test -- DO NOT MERGE

benjaminjb:


internal/cmd/create_test.go
```diff
@@ -59,3 +70,51 @@ spec:
...
+	}
+
+	err = postgresCluster.Run(context.TODO())
+	assert.NilError(t, err)
+
+	// list, err := drc.List(context.TODO(), metav1.ListOptions{})
+	// assert.NilError(t, err)
+	// log.Printf("list %s", list)
+
+	get, err := drc.Namespace("test").Get(context.TODO(), "hippo", metav1.GetOptions{})
```
cbandy, line 112: 🤔 Perhaps instead of going back through the API-call part, this should use the underlying [tracker](https://pkg.go.dev/k8s.io/client-go/dynamic/fake#FakeDynamicClient.Tracker)?

PR #79, opened 2023-10-31
Add timestamps to support

tony-landreth:
Issue: PGO-274


internal/cmd/exec.go
```diff
@@ -79,3 +79,9 @@ func (exec Executor) processes() (string, string, error) {
 
 	return stdout.String(), stderr.String(), err
 }
+
+func (exec Executor) systemTime() (string, string, error) {
+	var stdout, stderr bytes.Buffer
+	err := exec(nil, &stdout, &stderr, "bash", "-ceu", "--", "date")
```
cbandy, line 85: ❓ Is `date` a built-in of Bash? If not, can we skip that indirection?

📝 One of our pgAdmin images doesn't have Bash.

```suggestion
	err := exec(nil, &stdout, &stderr, "date")
```

internal/cmd/export.go
```diff
@@ -1078,6 +1084,101 @@ func gatherPatroniInfo(ctx context.Context,
...
+	}
+
+	podExec, err := util.NewPodExecutor(config)
+	if err != nil {
+		return err
+	}
+
+	var buf bytes.Buffer
+	for _, pod := range pods.Items {
+		for _, container := range pod.Spec.Containers {
```
cbandy, line 1123: 🤔 Once per pod should be plenty. Perhaps loop over the containers until the exec succeeds. RBAC also applies at the pod level.

Pseudocode:

```go
for range pods.Items {
  var stdout, stderr

  for range pod.Containers {
    stdout, stder, err = exec()
    if err == nil {
      break
    } else if IsForbidden(err) {
      writeInfo()
      break
    }
  }

  // diff, output, etc
}
```

internal/cmd/export.go
```diff
@@ -1078,6 +1084,101 @@ func gatherPatroniInfo(ctx context.Context,
...
+				continue
+			}
+
+			// Get client datetime.
+			clientDateTimeStr := time.Now().UTC().Format(time.UnixDate)
+			// Normalize client time for calculating delta.
+			clientTime, err := time.Parse(time.UnixDate, clientDateTimeStr)
+			if err != nil {
+				return err
+			}
```
cbandy, line 1150: ⛏️ I would do this in the other direction to avoid an error. Hold the time, then format it.

```suggestion
			clientTime := time.Now().UTC()
			clientDateTimeStr := clientTime.Format(time.UnixDate)
```

internal/cmd/export.go
```diff
@@ -1078,6 +1084,101 @@ func gatherPatroniInfo(ctx context.Context,
...
+
+			// Calculate difference between client and container datetime.
+			delta := clientTime.Sub(containerDateTime)
+			deltaStr := fmt.Sprintln(delta)
+
+			// Build report.
+			buf.Write([]byte("Pod: " + pod.GetName() + "\t" + "Container: " + container.Name + "\n"))
+			buf.Write([]byte("\tClient time: " + clientDateTimeStr + "\n"))
+			buf.Write([]byte("\tContainer time: " + containerDateTimeStr + "\n"))
+			buf.Write([]byte("\tDelta: " + deltaStr + "\n"))
```
cbandy, line 1167: 🤔 Parsing the text returned by `date` can error, and we do it only to show `Delta:`. Is it worth it? Dropping the calculation would simplify this area significantly.

internal/cmd/export.go
```diff
@@ -1078,6 +1084,101 @@ func gatherPatroniInfo(ctx context.Context,
...
+			}
+
+			// Calculate difference between client and container datetime.
+			delta := clientTime.Sub(containerDateTime)
+			deltaStr := fmt.Sprintln(delta)
+
+			// Build report.
+			buf.Write([]byte("Pod: " + pod.GetName() + "\t" + "Container: " + container.Name + "\n"))
+			buf.Write([]byte("\tClient time: " + clientDateTimeStr + "\n"))
+			buf.Write([]byte("\tContainer time: " + containerDateTimeStr + "\n"))
```
cbandy, line 1166: ⛏️ [fmt.Fprint](https://pkg.go.dev/fmt#Fprint) and friends are another way to write to a buffer.

```suggestion
			fmt.Fprintln(buf, "Pod: " + pod.GetName() + "\t" + "Container: " + container.Name))
			fmt.Fprintln(buf, "\tClient time: " + clientDateTimeStr))
			fmt.Fprintln(buf, "\tContainer time: " + containerDateTimeStr))
```

internal/cmd/export.go
tony-landreth: I should have made the comment clearer. The normalization here is to zero out time increments under a second. When you run `date` and then parse, milliseconds will be 00 because they're not included in the output: ```Wed Nov  1 07:07:17 AM EDT 2023```. 

When you run time.Now(), the returned value will include milliseconds and nanoseconds, observable with `time.Now().UTC().UnixMilli()`. If you don't set those values to zero, when you perform the subtraction you'll get sub second remainders that we don't care about... but maybe we don't care about taking the delta anyway.

internal/cmd/export.go
tony-landreth: Good point.

internal/cmd/export.go
```diff
@@ -1078,6 +1084,101 @@ func gatherPatroniInfo(ctx context.Context,
...
+			stdout, stderr, err := Executor(exec).systemTime()
+			if err != nil {
+				// If we get an RBAC error, let the user know. Otherwise, just
+				// try the next container.
+				if apierrors.IsForbidden(err) {
+					writeInfo(cmd, fmt.Sprintf(
+						"Failed to get system time for Container \"%s\" in Pod \"%s\". Error: \"%s\"",
+						container.Name, pod.GetName(), err.Error()))
+				}
+				continue
```
benjaminjb, line 1141: Will we be trying several containers in the same pod if we get a Forbidden error?

internal/cmd/export.go
benjaminjb:
```suggestion
			fmt.Fprintln(buf, "Pod: " + pod.GetName() + "\t" + "Container: " + container.Name + 
				"\tClient time: " + clientDateTimeStr + "\tContainer time: " + containerDateTimeStr)
```

internal/cmd/export.go
tony-landreth: Following Chris's advice, we can scale the exec back to checking once per pod.

internal/cmd/export.go
cbandy: That sounds like [Time.Truncate](https://pkg.go.dev/time#Time.Truncate). To ditch sub-second, I think you can call `t = t.Truncate(time.Second)`.

internal/cmd/exec.go
```diff
@@ -79,3 +79,9 @@ func (exec Executor) processes() (string, string, error) {
 
 	return stdout.String(), stderr.String(), err
 }
+
+func (exec Executor) systemTime() (string, string, error) {
```
tjmoore4, line 83: :pick: Maybe a comment here?

tjmoore4 (commented):
Looks good to me, but I'll hold off approving until the things in flight are resolved.

internal/cmd/export.go
tony-landreth: @benjaminjb Do you want everything on one line?

tony-landreth: This format might be easy to see what you're looking for at a glance:
```
Delta: 0s		 Pod name: hippo-instance1-9xl8-0
Pod time:		 Wed Nov  1 14:42:33 UTC 2023
Client time:		 Wed Nov  1 14:42:33 UTC 2023
```

benjaminjb: Just to verify, this is what the committed code currently does
```
Pod: hippo-instance1-9xl8-0   Container: database
    Client time: Wed Nov  1 14:42:33 UTC 2023
    Container time: Wed Nov  1 14:42:33 UTC 2023
    Delta: 0s
```
That's pretty clear about one pod's time, but I guess I was thinking that to compare times across pods, it might be easier if it were
```
Pod: hippo-instance1-9xl8-0   Container: database    Client time: Wed Nov  1 14:42:33 UTC 2023    Container time: Wed Nov  1 14:42:33 UTC 2023   Delta: 0s
Pod: hippo-repo-0   Container: pgbackrest    Client time: Wed Nov  1 14:42:33 UTC 2023    Container time: Wed Nov  1 14:42:33 UTC 2023   Delta: 0s
```
Of course, with pod and container being variable, the lines won't line up unless we pad or put them at the end (potentially time might also end up like that).

tony-landreth: That's correct about what's currently committed. And, good point about comparing pod times. Insofar as what you want to know first is temporal diffs,  maybe we should just lead with the times?
```
Delta: 0s	Pod time: Wed Nov  1 15:12:57 UTC 2023	Client time: Wed Nov  1 15:12:57 UTC 2023	Pod name: hippo-instance1-9xl8-0
Delta: 0s	Pod time: Wed Nov  1 15:12:57 UTC 2023	Client time: Wed Nov  1 15:12:57 UTC 2023	Pod name: hippo-instance1-v4jw-0
Delta: 0s	Pod time: Wed Nov  1 15:12:58 UTC 2023	Client time: Wed Nov  1 15:12:58 UTC 2023	Pod name: hippo-pgadmin-0
Delta: 0s	Pod time: Wed Nov  1 15:12:58 UTC 2023	Client time: Wed Nov  1 15:12:58 UTC 2023	Pod name: hippo-pgbouncer-6cc64db4d7-8prv7
Delta: 0s	Pod time: Wed Nov  1 15:12:59 UTC 2023	Client time: Wed Nov  1 15:12:59 UTC 2023	Pod name: hippo-repo-host-0
```

internal/cmd/export.go
tony-landreth: Yes. Are you thinking that if we get a Forbidden error, we should just move on to the next Pod?

benjaminjb: If we get a Forbidden on creating an exec to one particular container in one particular pod, would we ever get a different result from another container/another pod?

(Downside to trying like this: the possibility of a lot of logs about "Failed to get"; upside -- I guess there might be RBAC set up that allowed exec on some pods and not others. I'd be a little surprised if it was by container, but it might happen. I vote for leaving it for now.)

internal/cmd/export.go
```diff
@@ -1078,6 +1084,104 @@ func gatherPatroniInfo(ctx context.Context,
...
+	clientTime := time.Now().UTC()
+	clientDateTimeStr := clientTime.Format(time.UnixDate)
+
+	var deltaStr string
+	var containerDateTimeStr string
+	if containerDateTime, err := time.Parse(time.UnixDate, strings.TrimSpace(stdout)); err == nil {
+		// Calculate difference between client and container datetime.
+		containerDateTimeStr = containerDateTime.Format(time.UnixDate)
+		delta := clientTime.Sub(containerDateTime).Truncate(time.Second)
+		deltaStr = fmt.Sprint(delta)
```
benjaminjb, line 1167:
```suggestion
		deltaStr := fmt.Sprint(clientTime.Sub(containerDateTime).Truncate(time.Second))
```

internal/cmd/export.go
```diff
@@ -1078,6 +1084,104 @@ func gatherPatroniInfo(ctx context.Context,
...
+	} else {
+		// Parse failed, use stdout instead.
+		containerDateTimeStr = strings.TrimSpace(stdout)
+		deltaStr = "No result"
+	}
+
+	// Build report.
+	fmt.Fprintln(&buf, "Delta: "+deltaStr+"\tPod time: "+containerDateTimeStr+
+		"\tClient time: "+clientDateTimeStr+"\tPod name: "+pod.GetName())
```
benjaminjb, line 1176: Worth adding the container name?

benjaminjb (approved):
No blockers

internal/cmd/export.go
tony-landreth: I didn't want any pondering of why one container was appearing and not another.

internal/cmd/export.go
tony-landreth: Works for me.

internal/cmd/export.go
tony-landreth: It looks like we assumed RBAC on containers was possible [here](https://github.com/CrunchyData/postgres-operator-client/blob/main/internal/cmd/export.go#L1132)?

benjaminjb (approved):

internal/cmd/export.go
cbandy: There is no RBAC per-container that I know of. The ability to exec is on a pod and can be limited to specific pod names.

PR #80, opened 2023-11-02
Add suggestions to restore and backup

tony-landreth:
PGO-646

### DIFF
--- a/internal/cmd/backup.go
+++ b/internal/cmd/backup.go
@@ -16,6 +16,7 @@ package cmd
 
 import (
 	"context"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -128,6 +129,9 @@ postgresclusters/hippo backup initiated`)
 		)
 
 		if err != nil {
+			if strings.Contains(err.Error(), "conflict") {
+				cmd.Printf("SUGGESTION: The --force-conflicts flag may help in performing this operation.")
+			}
 			cmd.Printf("\nError requesting update: %s\n", err)
 			return err
 		}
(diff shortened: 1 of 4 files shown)


internal/cmd/backup.go
```diff
@@ -128,6 +129,9 @@ postgresclusters/hippo backup initiated`)
 		)
 
 		if err != nil {
+			if strings.Contains(err.Error(), "conflict") {
+				cmd.Printf("SUGGESTION: The --force-conflicts flag may help in performing this operation.")
```
benjaminjb, line 133: "\n" around printf?

benjaminjb (approved):
LGTM, no blockers, just want to make sure the output is well-spaced

internal/cmd/backup.go
tony-landreth: The error prints on the next line in this case.

internal/cmd/restore.go
```diff
@@ -190,6 +193,9 @@ func (config pgBackRestRestore) Run(ctx context.Context) error {
 		config.PostgresCluster, types.ApplyPatchType, patch,
 		config.Patch.PatchOptions(patchOptions))
 	if err != nil {
+		if strings.Contains(err.Error(), "conflict") {
```
cbandy, line 196: ⛏️ Perhaps this function identifies the error? https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsConflict

```suggestion
		if apierrors.IsConfict(err) {
```

benjaminjb: Hmm, yeah, makes sense to me:

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/managedfields/internal/conflict.go#L42
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/errors/errors.go#L249

tony-landreth: Nice! Thanks for the suggestion. Here's a [PR](https://github.com/CrunchyData/postgres-operator-client/pull/82).

PR #83, opened 2023-11-03
Add show user command

tjmoore4:
Adds a command to show the contents of the pguser Secrets.

Issue: PGO-470


internal/cmd/show.go
```diff
@@ -278,6 +283,143 @@ func showHA(
...
+		cmd.Print(stdout)
+
+		return nil
+	}
+
+	return cmdShowUser
+}
+
+// showUser returns a string with the decoded contents of the cluster's user Secrets.
+func showUser(config *internal.Config, args []string, showSensitive bool) (string, error) {
```
tjmoore4, line 339: :thought_balloon: Should this function be broken up? If so, what are your thoughts on where to do the break?

internal/cmd/show.go
```diff
@@ -278,6 +283,143 @@ func showHA(
...
+			b, err := yaml.Marshal(secret.Data[k])
+			if err != nil {
+				return output, err
+			}
+			d := make([]byte, base64.StdEncoding.EncodedLen(len(b)))
+			_, err = base64.StdEncoding.Decode(d, b)
+			if err != nil {
+				return output, err
+			}
+			if containsString(fields, k) {
```
tjmoore4, line 415: :thought_balloon: This currently lives in `delete.go`. Should it be moved, perhaps to `util`?

internal/cmd/show.go
```diff
@@ -278,6 +283,143 @@ func showHA(
...
+		fields = append(fields, sensitive...)
+
+		fmt.Print("WARNING: This command will show sensitive password information." +
+			"\nAre you sure you want to continue? (yes/no): ")
+
+		var confirmed *bool
+		for i := 0; confirmed == nil && i < 10; i++ {
+			// retry 10 times or until a confirmation is given or denied,
+			// whichever comes first
+			confirmed = confirm(os.Stdin, os.Stdout)
```
tjmoore4, line 357: :thought_balloon: This currently lives in `delete.go`. Should it be moved, perhaps to `util`?

internal/cmd/show.go
tony-landreth: Maybe broken up between getting and the presenting?

internal/cmd/show.go
tony-landreth: `util` sounds good.

internal/cmd/show.go
tony-landreth: Maybe we don't need that function anymore? There's `slices.Contains` starting in Go 1.21. Works like [this](https://go.dev/play/p/AIXKWIkLoN9).

internal/cmd/show.go
tjmoore4: :+1: I'll work on an update.

internal/cmd/show.go
tjmoore4: :thinking: Hmmm, I like the idea, but I don't know that we want to 1.21 (I know a lot of my environments have older versions). Maybe just move to `util` for now but add a TODO or something?

tjmoore4: hmmm, quick follow up. It looks like that function was added in Go [1.18](https://tip.golang.org/doc/go1.18). Maybe that's far enough back at this point?

tony-landreth: Sound reasonable to me.

internal/cmd/show.go
```diff
@@ -278,6 +285,150 @@ func showHA(
...
+	if err != nil {
+		return "", err
+	}
+
+	return getUserData(fields, list)
+}
+
+// getUserData returns the requested user data from the provided Secret List.
+// If the Secret List is empty, return a message stating that.
+func getUserData(fields []string, list *corev1.SecretList) (string, error) {
```
tony-landreth, line 395: :pick: The name is a little counter-intuitive to me, because of the `get` prefix -- makes me expect a GET request. Maybe just `userData`?

tony-landreth:
It doesn't look like the command works when I set the user through the spec:
```
spec:
  users:
    - name: rhino
      databases:
        - zoo
```

benjaminjb:
> It doesn't look like the command works when I set the user through the spec:
> 
> ```
> spec:
>   users:
>     - name: rhino
>       databases:
>         - zoo
> ```

What are you seeing?

If I set two users in the spec, one with the `databases` field filled out, like

```
    users:
    - name: hippo
      databases:
      - zoo
    - name: rhino
```

then the `show user` command just lists them (which is pretty boring now that I think of it):

```
➜ ./bin/kubectl-pgo show user hippo -n postgres-operator                       
CLUSTER  USERNAME
-------  --------
hippo    rhino
hippo    hippo
```

And then if I ask for the connection info, I get the databases[0] listed, as per the secret:
```
➜  ./bin/kubectl-pgo show user hippo hippo -n postgres-operator --show-connection-info
WARNING: This command will show sensitive password information.
Are you sure you want to continue? (yes/no): yes
Connection information for hippo for hippo cluster
Connection info string:
    dbname=zoo host=hippo-primary.postgres-operator.svc port=5432 user=hippo password=<password>
Connection URL:
    postgres://<password>a@hippo-primary.postgres-operator.svc:5432/zoo

**Comment by benjaminjb** [comment]
The `show user` without connection info is pretty boring, but I think we have some ideas abt making that command more like v4 in the future, i.e., pull info from the db.

**Comment by benjaminjb** [comment]
Current usage: Screen Shot 2023-11-09 at 2 44 54 PM

File: internal/cmd/show.go [hunk 5]
```diff
@@ -87,7 +92,7 @@ HA
 
 		// Print the pgbackrest info output received.
 		cmd.Printf("BACKUP\n\n")
-		if stdout, stderr, err := showBackup(config, args, "text", ""); err != nil {
+		if stdout, stderr, err := getBackup(config, args, "text", ""); err != nil {
```
tjmoore4, line 95: :pick: Not a blocker, but I'd like function renames unrelated to the current feature to be in a separate commit (same for `getHA` as well).

internal/cmd/show.go
```diff
@@ -278,6 +283,203 @@ func showHA(
...
+	clusterName string,
+) error {
+
+	// If the user is asking for connection strings, we use the alternate printer.
+	if showSensitive {
+		return printUserConnectionStrings(cmd, secretList, clusterName)
+	}
+
+	// Set up a tabwriter that writes to stdout, with
+
```
tjmoore4, line 415: :pick: extra newline

internal/cmd/show.go
```diff
@@ -278,6 +283,203 @@ func showHA(
...
+		if dbnameSet, ok := secret.Data["dbname"]; ok {
+			dbname = string(dbnameSet)
+		}
+		cmd.Println("    dbname=" + dbname +
+			" host=" + string(secret.Data["host"]) +
+			" port=" + string(secret.Data["port"]) +
+			" user=" + string(secret.Data["user"]) +
+			" password=" + string(secret.Data["password"]))
+		cmd.Println("Connection URL:")
+		cmd.Println("    postgres://" + string(secret.Data["password"]) +
```
tjmoore4, line 466: :question: Should we have a `user` value here? For a cluster named `hippo` (and based on this [example ](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS)), I'd expect something like 
```
Connection URL:
    postgres://user:password@hippo-primary.postgres-operator.svc:5432/hippo
```
So the code might be
```
cmd.Println("    postgres://" + string(secret.Data["user"]) + ":" + string(secret.Data["password"]) +
```

internal/cmd/show.go
benjaminjb: Pushed new commits to separate that.

internal/cmd/show.go
benjaminjb: Ooh, good catch

tjmoore4 (commented):
:+1: Updates look good to me.

PR #85, opened 2023-11-06
Adds kubectl plugin list to export

tony-landreth:
Issue: PGO-598


benjaminjb (commented):
LGTM, but I think we want to add documentation to the Help message re: what we're doing here with an exec.

docs/content/reference/pgo_support_export.md
```diff
@@ -68,7 +68,8 @@ kubectl pgo support export daisy --monitoring-namespace another-namespace --outp
 | PGO CLI Support Export Tool
 | The support export tool will collect information that is
 | commonly necessary for troubleshooting a PostgresCluster.
-| Note: No data or k8s secrets are collected.
+| Note: No data or k8s secrets are collected, however,
+| kubectl is used to list plugins on the user's machine.
```
tony-landreth, line 72: @benjaminjb Here's an attempt at documenting the exec.

internal/cmd/export.go
```diff
@@ -231,7 +232,8 @@ kubectl pgo support export daisy --monitoring-namespace another-namespace --outp
 | PGO CLI Support Export Tool
 | The support export tool will collect information that is
 | commonly necessary for troubleshooting a PostgresCluster.
-| Note: No data or k8s secrets are collected.
+| Note: No data or k8s secrets are collected, however,
+| kubectl is used to list plugins on the user's machine.
```
benjaminjb, line 236:
```suggestion
| Note: No data or k8s secrets are collected.
| However, kubectl is used to list kubectl plugins on the user's machine.
```
?
I just like the period after `No data or k8s secrets are collected.` It's a definitive statement.
Kubeectl to list kubectl plugins is redundant, but I want to be redundant here.

benjaminjb (approved):
LGTM, not a blocker, just a tweak to the language.

internal/cmd/export.go
tony-landreth: I think you're right. I was internally debating it.

benjaminjb (approved):

PR #86, opened 2023-11-06
Test matrix

benjaminjb:


tony-landreth (approved):

benjaminjb:
As much as this satisfies the symmetricist in me, gonna check out some deprecations/changes in the different versions so our testing spread satisfies.

benjaminjb:
OK, so v1.22 changes several API we want, then v1.25 did as well [cite](https://kubernetes.io/docs/reference/using-api/deprecation-guide/), so if we test 1.27, 1.24. and 1.21, we actually will hit all the changes.

benjaminjb:
I was also playing around with using GITHUB_ENV to set a matrix (thinking to randomize the midpoint test), but decided that I wanted this test to be more determinative in what it tested in the midrange.

PR #87, opened 2023-11-07
Separate backup

benjaminjb:
* separates out the backup logic from the cobra command
* adds a basic test to the backup command

Note: `backup` updates the cluster spec, which updates the job pod, so this command is best tested with an e2e test with a running PGO.

(The "cluster spec" is also unstructured, another reason why running an e2e test is the best way to ensure confidence.)

Issue: PGO-663

### DIFF
--- a/internal/cmd/backup.go
+++ b/internal/cmd/backup.go
@@ -92,7 +92,9 @@ postgresclusters/hippo backup initiated`)
 			cmd.Println(msg)
 		}
 		if err == nil {
-			// Consider a `--wait` flag
+			// Our `backup` command initiates a job, but does not signal to the user
+			// that a backup has finished; consider a `--wait` flag to wait until the
+			// backup is done.
 			cmd.Printf("%s/%s backup initiated\n", mapping.Resource.Resource, backup.ClusterName)
 		}
 
--- a/internal/cmd/backup.go
+++ b/internal/cmd/backup.go
@@ -69,9 +69,7 @@ postgresclusters/hippo backup initiated`)
 	// `backup` command accepts `repoName`, `force-conflicts` and `options` flags;
 	// multiple options flags can be used, with each becoming a new line
 	// in the options array on the spec
-	backup := pgBackRestBackup{
-		Config: config,
-	}
+	backup := pgBackRestBackupArgs{}
 	cmdBackup.Flags().BoolVar(&backup.ForceConflicts, "force-conflicts", false, "take ownership and overwrite the backup settings")
 	cmdBackup.Flags().StringVar(&backup.RepoName, "repoName", "", "repoName to backup to")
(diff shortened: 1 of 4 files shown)


testing/kuttl/e2e/backup/README.md
```diff
@@ -21,6 +21,7 @@
 * Check the annotation on the cluster
 * No backup occurs
 
-(6) 10-11
+(6) 10-12
 * Update the spec through KUTTL, changing the ownership of that field
 * Call the backup CLI with different flags, and see a conflict
+* Call the backup CLI with different flags and the `--force-conflict` flag, see no conflict
```
tony-landreth, line 27:
```suggestion
* Call the backup CLI with different flags and the `--force-conflicts` flag, see no conflict
```

internal/cmd/backup.go
```diff
@@ -185,3 +141,67 @@ func (config pgBackRestBackup) modifyIntent(
 
 	return nil
 }
+
+func (backup pgBackRestBackup) Run(client dynamic.NamespaceableResourceInterface,
+	cmd *cobra.Command,
+	clusterName string) error {
+
+	var (
```
benjaminjb, line 149: Remove

internal/cmd/backup.go
```diff
@@ -79,88 +79,45 @@ postgresclusters/hippo backup initiated`)
...
-		)
-
-		if err != nil {
-			if apierrors.IsConflict(err) {
-				cmd.Printf("SUGGESTION: The --force-conflicts flag may help in performing this operation.")
-			}
-			cmd.Printf("\nError requesting update: %s\n", err)
-			return err
+		if err == nil {
+			// Consider a `--wait` flag
```
tony-landreth, line 95: Not sure I understand this comment.

benjaminjb: When we request a backup, we initiate the process, but the user would then have to check if the backup job had a completed pod or check the spec to see that the most recent backup was complate. Would it be useful to wait until the backup is done?

benjaminjb: I should add that ^ context to the comment.

internal/cmd/backup.go
```diff
@@ -79,88 +79,47 @@ postgresclusters/hippo backup initiated`)
...
+type pgBackRestBackupArgs struct {
+	ClusterName    string
+	ForceConflicts bool
 	Options        []string
 	RepoName       string
-	ForceConflicts bool
 }
 
-func (config pgBackRestBackup) modifyIntent(
+func (backup pgBackRestBackupArgs) modifyIntent(
```
tony-landreth, line 114: Should `modifyIntent` methods be defined on Unstructured structs?

tony-landreth (approved):

PR #88, opened 2023-11-09
Adds start and stop commands

tony-landreth:
Issue: PGO-450

### DIFF
--- a/docs/content/reference/pgo_start.md
+++ b/docs/content/reference/pgo_start.md
@@ -1,20 +1,49 @@
 ---
-title: Command Reference
-aliases:
-- /reference/pgo
-weight: 100
+title: pgo start
 ---
-## pgo
+## pgo start
 
-pgo is a kubectl plugin for PGO, the open source Postgres Operator
+Start cluster
 
 ### Synopsis
 
-pgo is a kubectl plugin for PGO, the open source Postgres Operator from Crunchy Data.
+Start allows you to start a PostgreSQL cluster.
+The --force-conflicts flag may be required if the spec.shutdown field has been updated by another client.
 
-	https://github.com/CrunchyData/postgres-operator
+### RBAC Requirements
+    Resources                                           Verbs
+    ---------                                           -----
+    postgresclusters.postgres-operator.crunchydata.com  [get patch]
+
+### Usage
+
+```
+pgo start CLUSTER_NAME [flags]
+```
+
+### Examples
+
+```
+# Start a 'hippo' postgrescluster.
+pgo start hippo
+
+# Resolve ownership conflict
+pgo start hippo --force-conflicts
+
+```
+### Example output
+```
+postgresclusters/hippo backup initiated
+```
 
 ### Options
+
+```
+      --force-conflicts   take ownership and overwrite the shutdown setting
+  -h, --help              help for start
+```
+
+### Options inherited from parent commands
 
 ```
       --as string                      Username to impersonate for the operation. User could be a regular user or a service account in a namespace.
@@ -26,7 +55,6 @@ pgo is a kubectl plugin for PGO, the open source Postgres Operator from Crunchy 
       --client-key string              Path to a client key file for TLS
       --cluster string                 The name of the kubeconfig cluster to use
       --context string                 The name of the kubeconfig context to use
-  -h, --help                           help for pgo
       --insecure-skip-tls-verify       If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure
       --kubeconfig string              Path to the kubeconfig file to use for CLI requests.
   -n, --namespace string               If present, the namespace scope for this CLI request
@@ -39,13 +67,5 @@ pgo is a kubectl plugin for PGO, the open source Postgres Operator from Crunchy 
 
 ### SEE ALSO
 
-* [pgo backup](/reference/pgo_backup/)	 - Backup cluster
-* [pgo create](/reference/pgo_create/)	 - Create a resource
-* [pgo delete](/reference/pgo_delete/)	 - Delete a resource
-* [pgo restore](/reference/pgo_restore/)	 - Restore cluster
-* [pgo show](/reference/pgo_show/)	 - Show PostgresCluster details
-* [pgo start](/reference/pgo_start/)	 - Start cluster
-* [pgo stop](/reference/pgo_stop/)	 - Stop cluster
-* [pgo support](/reference/pgo_support/)	 - Crunchy Support commands for PGO
-* [pgo version](/reference/pgo_version/)	 - PGO client and operator versions
+* [pgo](/reference/)	 - pgo is a kubectl plugin for PGO, the open source Postgres Operator
 
--- a/docs/content/reference/pgo_start.md
+++ b/docs/content/reference/pgo_start.md
@@ -7,7 +7,7 @@ Start cluster
 
 ### Synopsis
 
-Start allows you to start a PostgreSQL cluster.
+Start sets the spec.shutdown field to false, allowing you to start a PostgreSQL cluster.
 The --force-conflicts flag may be required if the spec.shutdown field has been updated by another client.
 
 ### RBAC Requirements
@@ -33,7 +33,7 @@ pgo start hippo --force-conflicts
 ```
 ### Example output
 ```
-postgresclusters/hippo backup initiated
+postgresclusters/hippo start initiated
 ```
 
 ### Options
--- a/docs/content/reference/pgo_stop.md
+++ b/docs/content/reference/pgo_stop.md
@@ -7,7 +7,7 @@ Stop cluster
 
 ### Synopsis
 
-Stop allows you to stop a PostgreSQL cluster.
+Stop sets the spec.shutdown field to true, allowing you to stop a PostgreSQL cluster.
 The --force-conflicts flag may be required if the spec.shutdown field has been used before.
 
 ### RBAC Requirements
@@ -33,7 +33,7 @@ pgo stop hippo --force-conflicts
 ```
 ### Example output
 ```
-postgresclusters/hippo backup initiated
+postgresclusters/hippo stop initiated
 ```
 
 ### Options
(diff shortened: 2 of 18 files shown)


testing/kuttl/e2e/start-stop/03-assert.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: postgres-operator.crunchydata.com/v1beta1
```
tony-landreth, line 1: This isn't right. I'll fix it.

docs/content/reference/pgo_start.md
```diff
@@ -0,0 +1,71 @@
+---
+title: pgo start
+---
+## pgo start
+
+Start cluster
+
+### Synopsis
+
+Start allows you to start a PostgreSQL cluster.
```
benjaminjb, line 10: You could mention that it starts a cluster by setting the spec.shutdown field, since you bring it up later. (And this may help prevent confusion btw start & create?)

tony-landreth: Good idea :+1:

docs/content/reference/pgo_start.md
```diff
@@ -0,0 +1,71 @@
...
+# Start a 'hippo' postgrescluster.
+pgo start hippo
+
+# Resolve ownership conflict
+pgo start hippo --force-conflicts
+
+```
+### Example output
+```
+postgresclusters/hippo backup initiated
```
benjaminjb, line 36:
```suggestion
postgresclusters/hippo ? initiated
```

docs/content/reference/pgo_stop.md
```diff
@@ -0,0 +1,71 @@
+---
+title: pgo stop
+---
+## pgo stop
+
+Stop cluster
+
+### Synopsis
+
+Stop allows you to stop a PostgreSQL cluster.
```
benjaminjb, line 10: same comment as start

docs/content/reference/pgo_stop.md
```diff
@@ -0,0 +1,71 @@
...
+# Stop a 'hippo' postgrescluster.
+pgo stop hippo
+
+# Resolve ownership conflict
+pgo stop hippo --force-conflicts
+
+```
+### Example output
+```
+postgresclusters/hippo backup initiated
```
benjaminjb, line 36:
```suggestion
postgresclusters/hippo ? initiated
```

internal/cmd/delete_test.go
```diff
@@ -21,6 +21,8 @@ import (
 	"testing"
 
 	"gotest.tools/v3/assert"
+
+	"github.com/crunchydata/postgres-operator-client/internal/util"
 )
 
 func TestConfirmDelete(t *testing.T) {
```
benjaminjb, line 28: Maybe move this into `interactions_test.go` and rename `TestConfirm`?

docs/content/reference/pgo_start.md
tony-landreth: Looks like the example output is wrong in multiple ways. I'll adjust the print and the example.

internal/cmd/start.go
```diff
@@ -0,0 +1,159 @@
...
+	// Limit the number of args, that is, only one cluster name
+	cmdStart.Args = cobra.ExactArgs(1)
+
+	var forceConflicts bool
+	cmdStart.Flags().BoolVar(&forceConflicts, "force-conflicts", false, "take ownership and overwrite the shutdown setting")
+	cmdStart.RunE = func(cmd *cobra.Command, args []string) error {
+		mapping, client, err := v1beta1.NewPostgresClusterClient(config)
+		if err != nil {
+			fmt.Fprint(config.Out, err.Error())
+			return err
```
benjaminjb, line 76: We print the err here and then return the err -- does the parent func print the err it receives?

tony-landreth: Pretty sure it does. I'll get rid of this line.

internal/cmd/stop.go
```diff
@@ -0,0 +1,105 @@
...
+postgresclusters/hippo stop initiated`)
+
+	// Limit the number of args, that is, only one cluster name
+	cmdStop.Args = cobra.ExactArgs(1)
+
+	var forceConflicts bool
+	cmdStop.Flags().BoolVar(&forceConflicts, "force-conflicts", false, "take ownership and overwrite the shutdown setting")
+	cmdStop.RunE = func(cmd *cobra.Command, args []string) error {
+		fmt.Print("WARNING: Stopping a postgrescluster is not destructive but " +
+			"it will take your database offline until you run start. \nAre you sure you want " +
```
benjaminjb, line 59:
```suggestion
			"it will take your database offline until you restart it. \nAre you sure you want " +
```
?

internal/util/interactions_test.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"bytes"
+	"io"
+	"strings"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestConfirmDelete(t *testing.T) {
```
benjaminjb, line 26:
```suggestion
func TestConfirm(t *testing.T) {
```

testing/kuttl/e2e/start-stop/03-assert.yaml
```diff
@@ -0,0 +1,6 @@
...
+kind: PostgresCluster
+metadata:
+  name: start-stop-cluster
+spec:
+  shutdown: false
```
benjaminjb, line 6: I wouldn't mind a test that attempts to set it to something it's already set, to see that we get the DoNothing msg

tony-landreth: Good idea :+1:

internal/cmd/start.go
```diff
@@ -0,0 +1,160 @@
...
+func patchClusterShutdown(cluster *unstructured.Unstructured, client dynamic.NamespaceableResourceInterface, args ShutdownRequestArgs) error {
+	ctx := context.Background()
+
+	currShutdownVal, found, err := unstructured.NestedBool(cluster.Object, "spec", "shutdown")
+	if err != nil {
+		return err
+	}
+	// If the shutdown status is equal to the intent of the command, do nothing.
+	if found && currShutdownVal == args.NewShutdownValue {
+		fmt.Fprintf(args.Config.Out, args.DoNothingMsg)
```
benjaminjb, line 117: Is there a benefit to using `fmt.Fprintf(config.Out` vs. say, `cmd.Printf` (which I assume behind the scenes is calling something similar)?

[See](https://github.com/spf13/cobra/blob/a0a6ae020bb3899ff0276067863e50523f897370/command.go#L1397C13-L1397C28)

testing/kuttl/e2e/start-stop/03-assert.yaml
benjaminjb: (maybe that's a thing that could be done with a mock client in the future)

internal/cmd/start.go
```diff
@@ -0,0 +1,160 @@
...
+		args.Config.Patch.PatchOptions(patchOptions))
+	if err != nil {
+		if apierrors.IsConflict(err) {
+			fmt.Fprintf(args.Config.Out, "SUGGESTION: The --force-conflicts flag may help in performing this operation.\n")
+		}
+		return err
+	}
+	fmt.Fprintf(args.Config.Out, "%s/%s %s\n",
+		args.Mapping.Resource.Resource, args.ClusterName, args.InitiatedMsg)
+	return nil
```
benjaminjb, line 152: This function does a lot of printing. What would it look like if we returned `string, error`, and the calling function printed the message if it existed and returned the error?

benjaminjb: Or really, maybe our 3 states are:
1. error -- in which case we want to bubble the error up (and only in the case of the force-conflicts do we have an additional string to print)
2. success -- print success message
3. donothing -- print the do-nothing message

tony-landreth: That's a better way to go.

internal/cmd/start.go
tony-landreth: It's one less argument to the function.

internal/cmd/start.go
```diff
@@ -0,0 +1,160 @@
...
+	ClusterName      string
+	Config           *internal.Config
+	DoNothingMsg     string
+	ForceConflicts   bool
+	InitiatedMsg     string
+	Namespace        string
+	NewShutdownValue bool
+	Mapping          *meta.RESTMapping
+}
+
```
benjaminjb, line 43: 🤔 this struct has ~3 fields that are dependent, to wit:
- if my NewShutdownValue is true, my InitiatedMsg better be "stop initiated" and my DoNothingMsg better be "Cluster already Stopped."
- and for NewShutdownValue: false, "start initiated", "Cluster already Started"

So having them as separate fields allows them to drift when we wouldn't want them to. Would it make sense to remove them from the struct, and put logic in the func saying "when true, set to stop, when false, set to start"?

benjaminjb (approved):
LGTM, no blockers, just a thought

internal/cmd/start.go
tony-landreth: That's a good idea :+1:

internal/cmd/stop.go
```diff
@@ -0,0 +1,105 @@
...
+
+		mapping, client, err := v1beta1.NewPostgresClusterClient(config)
+		if err != nil {
+			fmt.Fprint(config.Out, err.Error())
+			return err
+		}
+		namespace, err := config.Namespace()
+		if err != nil {
+			fmt.Fprint(config.Out, err.Error())
+			return err
```
benjaminjb, line 78: Can just return the err, right? No need to print

internal/cmd/start.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+
+		cluster, err := getPostgresCluster(client, requestArgs)
+		if err != nil {
+			return err
+		}
+
+		msg, err := patchClusterShutdown(cluster, client, requestArgs)
+		if msg != "" {
+			cmd.Printf(msg)
```
benjaminjb, line 95: Since we're not formatting anything, maybe `cmd.Print` or `cmd.Println` and can skip the final newline in the messages?

benjaminjb (approved):

PR #92, opened 2023-12-01
Updated KUTTL E2Es for POSIX standard shells

tjmoore4:
Overall reference
- https://www.gnu.org/software/bash/manual/html_node/Major-Differences-From-The-Bourne-Shell.html

Some specific items include:

Double equals ('==') not supported by all POSIX shells.
- https://www.shellcheck.net/wiki/SC3014

printf %q not supported by all POSIX shells.
- https://ss64.com/bash/printf.html
- https://www.shellcheck.net/wiki/SC3050

arithmetic for loops are undefined
- https://www.shellcheck.net/wiki/SC3005

regex matching (=~) is undefined
- https://www.shellcheck.net/wiki/SC3015

cannot match globs in [..]
- https://www.shellcheck.net/wiki/SC2081

cannot use logical operators in [..]
- https://www.shellcheck.net/wiki/SC2107

Issue: PGO-716


testing/kuttl/e2e/backup/11--backup-with-new-flag.yaml
```diff
@@ -12,7 +12,14 @@ commands:
...
 
-    if [[ "${RESULT,,}" != *conflict* || "${RESULT}" != *repoName* ]]; then
-      printf 'Expected conflict on repoName, got %q\n' "${RESULT}"
-      exit 1
+    # get lowercase for result
+    lower=$(echo "$RESULT" | tr '[:upper:]' '[:lower:]')
+
+    # check for the substring using POSIX compliant parameter-expansion
+    # - https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html
+    if [ "${lower}" != "${lower%"conflict"*}" ] && [ "${RESULT}" != "${RESULT%"repoName"*}" ]; then
```
benjaminjb, line 20: What is going on here?

testing/kuttl/e2e/create/01--check_version_validation.yaml
```diff
@@ -4,43 +4,69 @@ kind: TestStep
...
+     esac
 
      # Verify the error when the Postgres version value is empty.
      NOT_SET=$(kubectl-pgo create postgrescluster --pg-major-version=  notset 2>&1)
-     if [[ "${NOT_SET}" != "Error: invalid argument"* ]]; then
-      printf 'Expected invalid argument error, got %q\n' "${NOT_SET}"
-      exit 1
-     fi
+     case "${NOT_SET}" in
+     "Error: invalid argument"*)
```
benjaminjb, line 66: `case` is interesting, do you prefer this to some of the `if` logic we have elsewhere?

testing/kuttl/e2e/restore/13--restore-specific.yaml
```diff
@@ -12,52 +12,57 @@ commands:
...
       STATUS=$(
         kubectl --namespace "${NAMESPACE}" get postgrescluster/restore-cluster \
           --output "jsonpath-as-json={.status.pgbackrest.restore.id}"
       )
 
-      [[ "${STATUS}" == "${RESTORE}" ]] && break
+      [ "${STATUS}" = "${RESTORE}" ] && break
 
       sleep 1
+    i=$(( i + 1 ))
```
benjaminjb, line 62: whitespace

testing/kuttl/e2e/create/01--check_version_validation.yaml
tjmoore4: The only reason I'm using case is to allow for 'contains' style checks, i.e. `if [ "${NOT_SET}" = "Error: invalid argument"* ]`. Written that way, it will fail because of the wildcard (see the [shellcheck issue](https://www.shellcheck.net/wiki/SC2081)) but you can use this type of check a `case ` statement. For any of the checks without wildcards, I left them as `if` statements.

testing/kuttl/e2e/backup/11--backup-with-new-flag.yaml
tjmoore4: The functionality should be the same as before, we are checking for the expected conflict.

Basically, if `lower`, which is a lower case version of `RESULT`, does not contain the string "conflict", it will return whatever `lower` contains, leaving lower = lower, which will fail the check. Similarly, `RESULT` is checked for the string "repoName". I used the parameter expansion method here instead of the `case` statement method I used elsewhere because I needed to simultaneously validate two different conditions. There may be a better option I'm missing, but this was a POSIX compliant option that met the requirements of the test.

tjmoore4: Updated the comment, per earlier discussion.

benjaminjb (approved):

testing/kuttl/e2e/show/01--pgbackrest-info.yaml
```diff
@@ -19,13 +19,13 @@ commands:
     )
 
     status=$?
-    if [ $status -ne 0 ]; then
+    if [ "$status" != 0 ]; then
```
dsessler7, line 22: Is there a particular reason why you changed this to `!=`? You've used `-ne` elsewhere.

dsessler7 (approved):
One nitpick. Otherwise, looks good!

testing/kuttl/e2e/show/01--pgbackrest-info.yaml
tjmoore4: That should be `-ne`, thanks for the catch!
`-ne` is for numeric checks while `!=` is for string comparisons ([reference](https://www.shellscript.sh/quickref.html)).

PR #102, opened 2024-08-14
More support export logs

philrhurst:
The pgo support export feature gathers Postgres logs from replicas in addition to the primary and pgbackrest logs where appropriate (in addition to those gathered in the container logs.

### DIFF
--- a/internal/cmd/export.go
+++ b/internal/cmd/export.go
@@ -961,9 +961,8 @@ func gatherAllPostgresLogs(ctx context.Context,
 ) error {
 	writeInfo(cmd, "Collecting Postgres logs...")
 
-	// The Primary Pod instance - there should be only one, but there may be none.
-	primaryPods, err := clientset.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{
-		LabelSelector: util.PrimaryInstanceLabels(clusterName),
+	dbPods, err := clientset.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: util.DBInstanceLabels(clusterName),
 	})
 
 	if err != nil {
@@ -974,36 +973,13 @@ func gatherAllPostgresLogs(ctx context.Context,
 		return err
 	}
 
-	if len(primaryPods.Items) == 0 {
-		writeDebug(cmd, "No primary instance pod found for gathering logs")
+	if len(dbPods.Items) == 0 {
+		writeDebug(cmd, "No database instance pod found for gathering logs")
 	}
 
-	writeDebug(cmd, fmt.Sprintf("Found %d Primary Pod\n", len(primaryPods.Items)))
+	writeDebug(cmd, fmt.Sprintf("Found %d Pods\n", len(dbPods.Items)))
 
-	// The Replica Pod instance - there may be zero or more.
-	replicaPods, err := clientset.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{
-		LabelSelector: util.ReplicaInstanceLabels(clusterName),
-	})
-
-	if err != nil {
-		if apierrors.IsForbidden(err) {
-			writeInfo(cmd, err.Error())
-			return nil
-		}
-		return err
-	}
-
-	writeDebug(cmd, fmt.Sprintf("Found %d Replica Pods\n", len(replicaPods.Items)))
-
-	var pods = primaryPods.Items
-	pods = append(pods, replicaPods.Items...)
-
-	if len(pods) == 0 {
-		writeDebug(cmd, "No pods found for gathering logs")
-		return nil
-	}
-
-	for _, pod := range pods {
+	for _, pod := range dbPods.Items {
 		writeDebug(cmd, fmt.Sprintf("Pod Name is %s\n", pod.Name))
 
 		podExec, err := util.NewPodExecutor(config)
@@ -1101,6 +1077,8 @@ func gatherAllPostgresLogs(ctx context.Context,
 	return nil
 }
 
+// gatherDbBackrestLogs gathers all the file-based pgBackRest logs on the DB instance.
+// There may not be any logs depending upon pgBackRest's log-level-file.
 func gatherDbBackrestLogs(ctx context.Context,
 	clientset *kubernetes.Clientset,
 	config *rest.Config,
@@ -1111,8 +1089,8 @@ func gatherDbBackrestLogs(ctx context.Context,
 ) error {
 	writeInfo(cmd, "Collecting pgBackRest logs...")
 
-	primaryPods, err := clientset.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{
-		LabelSelector: util.PrimaryInstanceLabels(clusterName),
+	dbPods, err := clientset.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: util.DBInstanceLabels(clusterName),
 	})
 
 	if err != nil {
@@ -1123,35 +1101,13 @@ func gatherDbBackrestLogs(ctx context.Context,
 		return err
 	}
 
-	if len(primaryPods.Items) == 0 {
-		writeDebug(cmd, "No primary instance pod found for gathering logs")
+	if len(dbPods.Items) == 0 {
+		writeDebug(cmd, "No database instance pod found for gathering logs")
 	}
 
-	writeDebug(cmd, fmt.Sprintf("Found %d Primary Pod\n", len(primaryPods.Items)))
+	writeDebug(cmd, fmt.Sprintf("Found %d Pods\n", len(dbPods.Items)))
 
-	replicaPods, err := clientset.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{
-		LabelSelector: util.ReplicaInstanceLabels(clusterName),
-	})
-
-	if err != nil {
-		if apierrors.IsForbidden(err) {
-			writeInfo(cmd, err.Error())
-			return nil
-		}
-		return err
-	}
-
-	writeDebug(cmd, fmt.Sprintf("Found %d Replica Pods\n", len(replicaPods.Items)))
-
-	var pods = primaryPods.Items
-	pods = append(pods, replicaPods.Items...)
-
-	if len(pods) == 0 {
-		writeDebug(cmd, "No pods found for gathering logs")
-		return nil
-	}
-
-	for _, pod := range pods {
+	for _, pod := range dbPods.Items {
 		writeDebug(cmd, fmt.Sprintf("Pod Name is %s\n", pod.Name))
 
 		podExec, err := util.NewPodExecutor(config)
@@ -1214,6 +1170,8 @@ func gatherDbBackrestLogs(ctx context.Context,
 	return nil
 }
 
+// gatherRepoHostLogs gathers all the file-based pgBackRest logs on the repo host.
+// There may not be any logs depending upon pgBackRest's log-level-file.
 func gatherRepoHostLogs(ctx context.Context,
 	clientset *kubernetes.Clientset,
 	config *rest.Config,
--- a/internal/cmd/export.go
+++ b/internal/cmd/export.go
@@ -446,12 +446,6 @@ Collecting PGO CLI logs...
 		}
 
 		// Logs
-		// if numLogs > 0 {
-		// 	if err == nil {
-		// 		err = gatherPostgresqlLogs(ctx, clientset, restConfig, namespace, clusterName, numLogs, tw, cmd)
-		// 	}
-		// }
-
 		// All Postgres Logs on the Postgres Instances (primary and replicas)
 		if numLogs > 0 {
 			if err == nil {
@@ -949,87 +943,6 @@ func gatherEvents(ctx context.Context,
 	path := clusterName + "/events"
 	if err := writeTar(tw, buf.Bytes(), path, cmd); err != nil {
 		return err
-	}
-
-	return nil
-}
-
-// gatherLogs takes a client and buffer to write logs to a buffer
-func gatherPostgresqlLogs(ctx context.Context,
-	clientset *kubernetes.Clientset,
-	config *rest.Config,
-	namespace string,
-	clusterName string,
-	numLogs int,
-	tw *tar.Writer,
-	cmd *cobra.Command,
-) error {
-	writeInfo(cmd, "Collecting Postgres logs...")
-	// Get the primary instance Pod by its labels
-	pods, err := clientset.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{
-		// TODO(jmckulk): should we be getting replica logs?
-		LabelSelector: util.PrimaryInstanceLabels(clusterName),
-	})
-	if err != nil {
-		if apierrors.IsForbidden(err) {
-			writeInfo(cmd, err.Error())
-			return nil
-		}
-		return err
-	}
-	if len(pods.Items) != 1 {
-		writeInfo(cmd, "No primary instance pod found for gathering logs")
-		return nil
-	}
-
-	podExec, err := util.NewPodExecutor(config)
-	if err != nil {
-		return err
-	}
-
-	exec := func(stdin io.Reader, stdout, stderr io.Writer, command ...string,
-	) error {
-		return podExec(namespace, pods.Items[0].GetName(), util.ContainerDatabase,
-			stdin, stdout, stderr, command...)
-	}
-
-	stdout, stderr, err := Executor(exec).listPGLogFiles(numLogs)
-	if err != nil {
-		if apierrors.IsForbidden(err) {
-			writeInfo(cmd, err.Error())
-			return nil
-		}
-		return err
-	}
-	if stderr != "" {
-		writeInfo(cmd, stderr)
-	}
-
-	logFiles := strings.Split(strings.TrimSpace(stdout), "\n")
-	for _, logFile := range logFiles {
-		var buf bytes.Buffer
-
-		stdout, stderr, err := Executor(exec).catFile(logFile)
-		if err != nil {
-			if apierrors.IsForbidden(err) {
-				writeInfo(cmd, err.Error())
-				// Continue and output errors for each log file
-				// Allow the user to see and address all issues at once
-				continue
-			}
-			return err
-		}
-
-		buf.Write([]byte(stdout))
-		if stderr != "" {
-			str := fmt.Sprintf("\nError returned: %s\n", stderr)
-			buf.Write([]byte(str))
-		}
-
-		path := clusterName + "/logs/postgresql/" + logFile
-		if err := writeTar(tw, buf.Bytes(), path, cmd); err != nil {
-			return err
-		}
 	}
 
 	return nil
--- a/internal/cmd/export.go
+++ b/internal/cmd/export.go
@@ -997,8 +997,10 @@ func gatherPostgresLogsAndConfigs(ctx context.Context,
 		// Get Postgres Log Files
 		stdout, stderr, err := Executor(exec).listPGLogFiles(numLogs)
 
-		// Depending upon the list* function above, err may be non-nil or stderr may
-		// be non-empty, indicating an error has happened.
+		// Depending upon the list* function above:
+		// An error may happen when err is non-nil or stderr is non-empty.
+		// In both cases, we want to print helpful information and continue to the
+		// next iteration.
 		if err != nil || stderr != "" {
 
 			if apierrors.IsForbidden(err) {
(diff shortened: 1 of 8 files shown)


internal/cmd/export.go
```diff
@@ -948,73 +960,349 @@ func gatherPostgresqlLogs(ctx context.Context,
...
 		}
 		return err
 	}
-	if len(pods.Items) != 1 {
-		writeInfo(cmd, "No primary instance pod found for gathering logs")
+
+	writeDebug(cmd, fmt.Sprintf("Found %d Replica Pods\n", len(replicaPods.Items)))
+
+	var pods = primaryPods.Items
+	pods = append(pods, replicaPods.Items...)
```
andrewlecuyer, line 999: 🤔 Instead of making separate/individual API requests for the primary and replica pods and appending here, can a single request be used that just grabs all of the instances in one go?  E.g. similar to the following `kubectl` command:

```bash
$ kubectl get pods -n postgres-operator --selector postgres-operator.crunchydata.com/cluster,postgres-operator.crunchydata.com/instance
NAME                     READY   STATUS    RESTARTS   AGE
hippo-instance1-48gt-0   4/4     Running   0          15h
hippo-instance1-vkhd-0   4/4     Running   0          15h
```

You could then just iterate on the items/instances from that single request.

internal/util/naming.go
```diff
@@ -69,6 +78,19 @@ func PrimaryInstanceLabels(clusterName string) string {
...
+		LabelData + "=" + DataPostgres + "," +
+		LabelRole + "=" + RolePatroniReplica
+}
+
+// RepoHostInstanceLabels provides labels for a Backrest Repo Host instances
+func RepoHostInstanceLabels(clusterName string) string {
+	return LabelCluster + "=" + clusterName + "," +
+		LabelData + "=" + DataBackrest
+}
+
```
andrewlecuyer, line 93: 🔧 The dedicated repo host is also assigned the `pgbackrest-dedicated` label.

https://github.com/CrunchyData/postgres-operator/blob/20cc36d5d60cc7b49057fe3d7262d57f8cf51b17/internal/naming/labels.go#L65

```bash
$ kubectl get pods -n postgres-operator --selector postgres-operator.crunchydata.com/cluster,postgres-operator.crunchydata.com/pgbackrest-dedicated
NAME                READY   STATUS    RESTARTS   AGE
hippo-repo-host-0   2/2     Running   0          15h
```

internal/cmd/export.go
```diff
@@ -948,73 +960,349 @@ func gatherPostgresqlLogs(ctx context.Context,
...
+			if err := writeTar(tw, buf.Bytes(), path, cmd); err != nil {
+				return err
+			}
+		}
+
+	}
+	return nil
+}
+
+func gatherDbBackrestLogs(ctx context.Context,
```
andrewlecuyer, line 1104: ⛏️ A quick comment describing this function is appreciated.  Same for function `gatherRepoHostLogs()` below.

internal/cmd/export.go
philrhurst: Good idea - I'll refactor. This is also happening in the `gatherDbBackrestLogs` function.

internal/util/naming.go
philrhurst: Do you prefer selecting on the `pgbackrest-dedicated` label?

I was selecting the Postgres instances with `postgres-operator.crunchydata.com/data=postgres` and the Backrest repo host with `postgres-operator.crunchydata.com/data=pgbackrest`. I like the similarity there; but will defer.

internal/cmd/export.go
philrhurst: I've added a comment to both functions.

internal/util/naming.go
philrhurst: I've refactored to use the pgbackrest-dedicates rather than data=pgbackrest.

caitlinstrong:
I tested this in my local environment (CPK 5.6.0, PG 16) and confirmed: 

- Postgres log files are retrieved from the primary and replica db instances 
- Postgres conf files are retrieved from the primary and replica db instances 
- pgBackRest log files are retrieved from the primary and replica db instances 
- pgBackRest log files are retrieved from the repo host
- the 'log' directory contains a folder for each pod in the cluster.  Each pod folder has a 'container' folder for container logs, and a 'pgdata' folder for the Postgres logs/conf files and pgBackRest logs: 

```
📦logs
 ┣ 📂hippo-backup-9bnh-4shxp
 ┃ ┗ 📂containers
 ┃ ┃ ┗ 📜pgbackrest.log
 ┣ 📂hippo-instance1-lqd7-0 (REPLICA)
 ┃ ┣ 📂containers
 ┃ ┃ ┣ 📜database.log
 ┃ ┃ ┣ 📜nss-wrapper-init.log
 ┃ ┃ ┣ 📜pgbackrest-config.log
 ┃ ┃ ┣ 📜pgbackrest.log
 ┃ ┃ ┣ 📜postgres-startup.log
 ┃ ┃ ┗ 📜replication-cert-copy.log
 ┃ ┗ 📂pgdata
 ┃ ┃ ┣ 📂pg16
 ┃ ┃ ┃ ┣ 📂log
 ┃ ┃ ┃ ┃ ┗ 📜postgresql-Tue.log
 ┃ ┃ ┃ ┣ 📜pg_hba.conf
 ┃ ┃ ┃ ┣ 📜pg_ident.conf
 ┃ ┃ ┃ ┣ 📜postgresql.auto.conf
 ┃ ┃ ┃ ┣ 📜postgresql.base.conf
 ┃ ┃ ┃ ┗ 📜postgresql.conf
 ┃ ┃ ┗ 📂pgbackrest
 ┃ ┃ ┃ ┗ 📂log
 ┃ ┃ ┃ ┃ ┗ 📜db-restore.log
 ┣ 📂hippo-instance1-zftn-0 (PRIMARY)
 ┃ ┣ 📂containers
 ┃ ┃ ┣ 📜database.log
 ┃ ┃ ┣ 📜nss-wrapper-init.log
 ┃ ┃ ┣ 📜pgbackrest-config.log
 ┃ ┃ ┣ 📜pgbackrest.log
 ┃ ┃ ┣ 📜postgres-startup.log
 ┃ ┃ ┗ 📜replication-cert-copy.log
 ┃ ┗ 📂pgdata
 ┃ ┃ ┣ 📂pg16
 ┃ ┃ ┃ ┣ 📂log
 ┃ ┃ ┃ ┃ ┗ 📜postgresql-Tue.log
 ┃ ┃ ┃ ┣ 📜pg_hba.conf
 ┃ ┃ ┃ ┣ 📜pg_ident.conf
 ┃ ┃ ┃ ┣ 📜postgresql.auto.conf
 ┃ ┃ ┃ ┣ 📜postgresql.base.conf
 ┃ ┃ ┃ ┗ 📜postgresql.conf
 ┃ ┃ ┗ 📂pgbackrest
 ┃ ┃ ┃ ┗ 📂log
 ┃ ┃ ┃ ┃ ┗ 📜db-stanza-create.log
 ┣ 📂hippo-repo-host-0
 ┃ ┣ 📂containers
 ┃ ┃ ┣ 📜nss-wrapper-init.log
 ┃ ┃ ┣ 📜pgbackrest-config.log
 ┃ ┃ ┣ 📜pgbackrest-log-dir.log
 ┃ ┃ ┗ 📜pgbackrest.log
 ┃ ┗ 📂pgbackrest
 ┃ ┃ ┗ 📂repo1
 ┃ ┃ ┃ ┗ 📂log
 ┃ ┃ ┃ ┃ ┣ 📜db-backup.log
 ┃ ┃ ┃ ┃ ┗ 📜db-expire.log
 ┗ 📜cli
```

benjaminjb:
I'm curious, what do you get if you run this command while a backup job is running for the cluster?

internal/cmd/export.go
```diff
@@ -446,12 +446,23 @@ Collecting PGO CLI logs...
 		}
 
 		// Logs
+		// All Postgres Logs on the Postgres Instances (primary and replicas)
 		if numLogs > 0 {
```
benjaminjb, line 450: Would we ever want similar logic for the other logs we're getting? Or are postgres logs different because they are more verbose (so we want to limit the number of them we get)?

internal/cmd/export.go
```diff
@@ -948,73 +960,307 @@ func gatherPostgresqlLogs(ctx context.Context,
...
 			return nil
 		}
 		return err
 	}
-	if len(pods.Items) != 1 {
-		writeInfo(cmd, "No primary instance pod found for gathering logs")
-		return nil
+
+	if len(dbPods.Items) == 0 {
+		writeDebug(cmd, "No database instance pod found for gathering logs")
```
benjaminjb, line 977: `writeDebug` only writes to the CLI log, so if someone runs this and there's no pods, they won't see this message in their stdout, right?

internal/cmd/export.go
```diff
@@ -948,73 +960,307 @@ func gatherPostgresqlLogs(ctx context.Context,
...
+				return err
+			}
+
+			buf.Write([]byte(stdout))
+			if stderr != "" {
+				str := fmt.Sprintf("\nError returned: %s\n", stderr)
+				buf.Write([]byte(str))
+			}
+
+			path := clusterName + fmt.Sprintf("/logs/%s/", pod.Name) + logFile
```
benjaminjb, line 1070: We're putting the confs in a directory called logs?

internal/cmd/export.go
```diff
@@ -937,8 +948,9 @@ func gatherEvents(ctx context.Context,
 	return nil
 }
 
-// gatherLogs takes a client and buffer to write logs to a buffer
-func gatherPostgresqlLogs(ctx context.Context,
+// gatherAllPostgresLogs take a client and writes logs from primary and replicas
+// to a buffer
+func gatherAllPostgresLogs(ctx context.Context,
```
benjaminjb, line 953:
```suggestion
func gatherPostgresLogsAndConfigs(ctx context.Context,
```

To me "gatherAllPostgresLogs" sounds like you're getting all the logs, but does not indicate where those logs are coming from. "gatherPostgresLogsFromAllDatabasePods" might be more indicative of the action you're performing here, but it's a little wordy.

But then what we're really doing here is getting the logs and the confs, so I wanted to make sure we capture that idea in the name.

I'm not sure what the best name here is that balances out "full description" and "not too many words for the sake of legibility".

internal/cmd/export.go
philrhurst: I don't think we would want to limit this for other logs. The case for limiting the PG logs is simply their size.

internal/cmd/export.go
philrhurst: correct - I see a need to change it to `writeInfo` - I'll make that change.

internal/cmd/export.go
philrhurst: the logs directory is serving some diverse needs right now. In this PR it contains the following kinds of data, all grouped by the originating Pod.

1. Postgres logs (file-based)
2. pgBackRest logs (file-based)
3. Postgres *.conf files
4. Container logs

I think overall the Support Export is getting different kinds of information. 

1. Info about K8s resources - this is usually the YAML representation of the resource
2. Container logs
3. Processes within the containers
4. Postgres-specific information (logs and settings)
5. Patroni-specific information (patronictl output)
6. BackRest-specific information (pgbackrest info and, hopefully soon, pgbackrest check)

Figuring out how to represent that externally is a challenge, but I think there's a natural order that can be identified. Probably grouping by Pod is the best since that's a basic unit of work within k8s. But I also avoid too much change at once. So I shoved the PG settings into logs because Postgres logs and Postgres setting are closely related.

internal/cmd/export.go
philrhurst: I like `gatherPostgresLogsAndConfigs` - I originally just added `All` because I was getting logs from the primary _and_ the replicas. So it went from gatherPG to gatherAllPG.

I'll refactor to `gatherPostgresLogsAndConfigs`

internal/cmd/export.go
benjaminjb: I haven't looked at this command too deeply in a while, so please don't take me as an authority; and I definitely see some challenges in organizing the data we capture. 

But can we change this path to 
```suggestion
			path := clusterName + fmt.Sprintf("/conf/%s/", pod.Name) + logFile
```
just for the conf files?

philrhurst: I think maybe changing the path from `/logs/...` to `/logs-and-config/...` may capture the spirit here. Having too flat of a structure makes it harder to examine the data within each pod. That's the challenge I'm facing here.

philrhurst: ideally, I'd rather use `/pods/...` but we already have a `pods` directory that holds the YAML for each Pod. I didn't want this PR to touch too many other things.... but I'm open to suggestions.

benjaminjb: I'm trying to brainstorm to see if there's a path I like more than ~`pods/<name>/logs-and-config/...` and there really isn't. What's in this directory? Logs and config files. Where do they come from? The pod named <name>.

Sounds good to me and we can always iterate if we think of something more descriptive. Thanks!

internal/cmd/export.go
```diff
@@ -948,73 +960,307 @@ func gatherPostgresqlLogs(ctx context.Context,
...
+			return podExec(namespace, pod.Name, util.ContainerDatabase,
+				stdin, stdout, stderr, command...)
+		}
+
+		stdout, stderr, err := Executor(exec).listBackrestLogFiles()
+
+		if err != nil {
+			writeInfo(cmd, err.Error())
+			if strings.Contains(stderr, "No such file or directory") {
+				writeDebug(cmd, "Cannot find any Backrest log files. This is acceptable in some configurations.\n")
```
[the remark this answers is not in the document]
philrhurst: Yes but I found some places when a `continue` would help the CLI recover and still gather. I'll push those improvements in a bit.

philrhurst:
@benjaminjb 

> I'm curious, what do you get if you run this command while a backup job is running for the cluster?

For this one - the CLI is pulling the Pod logs from the Pod created by the CronJob. Depending upon the backrest config, there can be a lot of detail there. If `log-level-file` is turned on, then the backrest repo Pod has logs for the backup (I don't think the CronJob will log anything to file). The CLI will pull whatever has been written to the file whenever the CLI runs that function.

internal/cmd/export.go
philrhurst: I may have found something palatable. I went with `pods`. So now there is no top-level `logs` folder. All the file-based logs roll up under the `pods` directory with the container name as a subdir. I like that when viewing each Pods' subdir, each file has the same relative path as within the Pod. It makes it easier to reason about.

These logs/conf files follow their on-disk structure (i.e. within `pgdata` or equivalent)

- PG Log files: `./pods/<container-name>/pgdata/pgNN/log/`
- PG Conf files: `./pods/<container-name>/pgdata/pgNN/`
- DB Backrest Log files: `./pods/<container-name>/pgdata/pgbackrest/`
- Repo Host Log files: `./pods/<container-name>/pgbackrest/repoN/log`

The container logs all follow a similar structure: 

- Container logs: `./pods/<container-name>/containers/`

I moved the cli log file to the root dir and gave it an extension: `/crunchy_k8s_support_export_2024-08-30-190004+0000/hippo/cli.log`. The extension is nice when viewing in VSCode.

internal/cmd/export.go
```diff
@@ -937,8 +948,9 @@ func gatherEvents(ctx context.Context,
 	return nil
 }
 
-// gatherLogs takes a client and buffer to write logs to a buffer
-func gatherPostgresqlLogs(ctx context.Context,
+// gatherPostgresLogsAndConfigs take a client and writes logs from primary and replicas
+// to a buffer
```
benjaminjb, line 952:
```suggestion
// gatherPostgresLogsAndConfigs take a client and writes logs and configs
// from primary and replicas to a buffer
```

internal/cmd/export.go
```diff
@@ -948,73 +960,313 @@ func gatherPostgresqlLogs(ctx context.Context,
...
+			if apierrors.IsForbidden(err) {
+				writeInfo(cmd, err.Error())
+				return nil
+			}
+			continue
+		}
+		if stderr != "" {
+			writeDebug(cmd, stderr)
+			continue
+		}
```
benjaminjb, line 1012:
```suggestion
		if err != nil {
			if apierrors.IsForbidden(err) {
				writeInfo(cmd, err.Error())
				return nil
			}
			continue
		}
		if stderr != "" {
			if strings.Contains(stderr, "No such file or directory") {
				writeDebug(cmd, "Cannot find any Postgres log files. This is acceptable in some configurations.\n")
			} else {
				writeDebug(cmd, stderr)
			}
			continue
		}
```
What do you think about moving the stderr check into its own discrete section?

internal/cmd/export.go
```diff
@@ -948,73 +960,313 @@ func gatherPostgresqlLogs(ctx context.Context,
...
+		writeInfo(cmd, "No database instance pod found for gathering logs")
+		return nil
+	}
+
+	writeDebug(cmd, fmt.Sprintf("Found %d Pods\n", len(dbPods.Items)))
+
+	for _, pod := range dbPods.Items {
+		writeDebug(cmd, fmt.Sprintf("Pod Name is %s\n", pod.Name))
+
+		podExec, err := util.NewPodExecutor(config)
```
benjaminjb, line 1120: 🤔 
Could you declare this NewPodExecutor outside of the loop and then reuse it on every loop?

internal/cmd/export.go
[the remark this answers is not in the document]
benjaminjb: Sorry I missed this before, but I think this stderr check could go below in the stderr block (lines 1143-1146 currently)

internal/cmd/export.go
```diff
@@ -948,73 +960,313 @@ func gatherPostgresqlLogs(ctx context.Context,
...
+			return podExec(namespace, pod.Name, util.ContainerPGBackrest,
+				stdin, stdout, stderr, command...)
+		}
+
+		// Get BackRest Repo Host Log Files
+		stdout, stderr, err := Executor(exec).listBackrestRepoHostLogFiles()
 		if err != nil {
+			if strings.Contains(stderr, "No such file or directory") {
+				writeDebug(cmd, "Cannot find any Repo Host log files. This is acceptable in some configurations.\n")
+			}
```
benjaminjb, line 1229: Ditto.

internal/cmd/export.go
philrhurst: updated and committed

internal/cmd/export.go
philrhurst: I think I can simplify this further if I can assume that any error would populate `err` and `stderr`. Is that a safe assumption in this case?

philrhurst: Basically reduce that section to the following.

My overall goal is just to emit a helpful message when no PG logs are found, but still `continue` or return nil` when appropriate.

```
		// Get Postgres Log Files
		stdout, stderr, err := Executor(exec).listPGLogFiles(numLogs)
		if err != nil {
			if strings.Contains(stderr, "No such file or directory") {
				writeDebug(cmd, "Cannot find any Postgres log files. This is acceptable in some configurations.\n")
			}
			if apierrors.IsForbidden(err) {
				writeInfo(cmd, err.Error())
				return nil
			}
			writeDebug(cmd, stderr)
			continue
		}
```

benjaminjb: > I think I can simplify this further if I can assume that any error would populate err and stderr. Is that a safe assumption in this case?

I don't know if that's a safe assumption. The way I think about it, "stderr" is when the function returns an error, while "err" is the possibility that the function doesn't even run.

Some code for an executor: https://github.com/CrunchyData/postgres-operator-client/blob/9782c9b2d75a5cfbbc7992197bbd3199cb5437c9/internal/util/executor.go#L56-L64

For instance, if you run the command to find files, and there are no files, we get the stderr: "no files". But if you run the command to find files and Kubernetes has deleted your pod, we get an err telling us there's no pod with that name.

internal/cmd/export.go
```diff
@@ -948,73 +960,365 @@ func gatherPostgresqlLogs(ctx context.Context,
...
+
+			if err != nil {
+				writeDebug(cmd, fmt.Sprintf("%s\n", err.Error()))
+			}
+			if stderr != "" {
+				writeDebug(cmd, stderr)
+			}
+
+			if strings.Contains(stderr, "No such file or directory") {
+				writeDebug(cmd, "Cannot find any pgBackRest log files. This is acceptable in some configurations.\n")
```
benjaminjb, line 1288: ⛏️ This function is really to make your life easier, so you know better than me, but these logs in `gatherRepoHostLogs` look the same as the logs in `gatherDbBackrestLogs` -- will that be confusing? It doesn't look that confusing to me, but wanted to check.

benjaminjb (approved):
Looks good to me, just one question about a line.

internal/cmd/export.go
philrhurst: they're different logs and coming from different paths - they could be abstracted into a common function (along with the PG logs), but right now they are very different.

PR #105, opened 2024-09-24
Large postgres logs

philrhurst:
gathering large Postgres files may pose a challenge for some systems. This change stores the file on disk, compares the filesize with the remote, and adds the local file to the tar. It has error handling to allow a support export to be generated when there are issues with gathering PG logs.

### DIFF
--- a/internal/cmd/export.go
+++ b/internal/cmd/export.go
@@ -49,7 +49,7 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
-	remotecommand "k8s.io/client-go/tools/remotecommand"
+	"k8s.io/client-go/tools/remotecommand"
 	"sigs.k8s.io/yaml"
 
 	"github.com/crunchydata/postgres-operator-client/internal"
@@ -1042,7 +1042,7 @@ func gatherPostgresLogsAndConfigs(ctx context.Context,
 
 		for _, logFile := range logFiles {
 			// get the file size to stream
-			fileSize, err := getRemoteFileSize(clientset, config, namespace, pod.Name, util.ContainerDatabase, logFile)
+			fileSize, err := getRemoteFileSize(config, namespace, pod.Name, util.ContainerDatabase, logFile)
 			if err != nil {
 				writeDebug(cmd, fmt.Sprintf("could not get file size for %s: %v\n", logFile, err))
 				continue
@@ -1951,44 +1951,28 @@ func addFileToTar(tw *tar.Writer, localPath, tarPath string) error {
 }
 
 // getRemoteFileSize returns the size of a file within a container so that we can stream its contents
-func getRemoteFileSize(clientset *kubernetes.Clientset, config *rest.Config,
+func getRemoteFileSize(config *rest.Config,
 	namespace string, podName string, containerName string, filePath string) (int64, error) {
 
-	// Prepare the command to get the file size using "stat -c %s <file>"
-	req := clientset.
-		CoreV1().
-		RESTClient().
-		Post().
-		Namespace(namespace).
-		Resource("pods").
-		Name(podName).
-		SubResource("exec").
-		Param("container", containerName).
-		Param("command", "stat").
-		Param("command", "-c").
-		Param("command", "%s").
-		Param("command", filePath).
-		Param("stderr", "true").
-		Param("stdout", "true")
-
-	exec, err := remotecommand.NewSPDYExecutor(config, "GET", req.URL())
+	podExec, err := util.NewPodExecutor(config)
 	if err != nil {
-		return 0, fmt.Errorf("could not initialize SPDY executor for stat: %w", err)
+		return 0, fmt.Errorf("could not create executor: %w", err)
+	}
+	exec := func(stdin io.Reader, stdout, stderr io.Writer, command ...string,
+	) error {
+		return podExec(namespace, podName, containerName,
+			stdin, stdout, stderr, command...)
 	}
 
-	var stdout, stderr strings.Builder
-	err = exec.Stream(remotecommand.StreamOptions{
-		Stdin:  nil,
-		Stdout: &stdout,
-		Stderr: &stderr,
-		Tty:    false,
-	})
+	// Prepare the command to get the file size using "stat -c %s <file>"
+	command := fmt.Sprintf("stat -c %s %s", "%s", filePath)
+	stdout, stderr, err := Executor(exec).bashCommand(command)
 	if err != nil {
-		return 0, fmt.Errorf("could not get file size: %w, stderr: %s", err, stderr.String())
+		return 0, fmt.Errorf("could not get file size: %w, stderr: %s", err, stderr)
 	}
 
 	// Parse the file size from stdout
-	size, err := strconv.ParseInt(strings.TrimSpace(stdout.String()), 10, 64)
+	size, err := strconv.ParseInt(strings.TrimSpace(stdout), 10, 64)
 	if err != nil {
 		return 0, fmt.Errorf("failed to parse file size: %w", err)
 	}
--- a/internal/cmd/export.go
+++ b/internal/cmd/export.go
@@ -1849,14 +1849,21 @@ func streamFileFromPod(clientset *kubernetes.Clientset,
 
 	// create localPath to write the streamed data from remotePath
 	localPath := filepath.Join(outputDir, filepath.Base(remotePath))
-	if err := os.MkdirAll(filepath.Dir(localPath), 0770); err != nil {
+	if err := os.MkdirAll(filepath.Dir(localPath), 0750); err != nil {
 		return fmt.Errorf("failed to create path for file: %w", err)
 	}
-	outFile, err := os.Create(localPath)
+	outFile, err := os.Create(filepath.Clean(localPath))
 	if err != nil {
 		return fmt.Errorf("failed to create local file: %w", err)
 	}
-	defer outFile.Close()
+
+	defer func() {
+		// ignore any errors from Close functions, the writers will be
+		// closed when the program exits
+		if outFile != nil {
+			_ = outFile.Close()
+		}
+	}()
 
 	// request to cat remotePath
 	req := clientset.CoreV1().RESTClient().
@@ -1921,11 +1928,17 @@ func streamFileFromPod(clientset *kubernetes.Clientset,
 // addFileToTar copies a local file into a tar archive
 func addFileToTar(tw *tar.Writer, localPath, tarPath string) error {
 	// Open the file to be added to the tar
-	file, err := os.Open(localPath)
+	file, err := os.Open(filepath.Clean(localPath))
 	if err != nil {
 		return fmt.Errorf("failed to open file: %w", err)
 	}
-	defer file.Close()
+	defer func() {
+		// ignore any errors from Close functions, the writers will be
+		// closed when the program exits
+		if file != nil {
+			_ = file.Close()
+		}
+	}()
 
 	// Get file info to create tar header
 	fileInfo, err := file.Stat()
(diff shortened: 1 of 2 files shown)


internal/cmd/export.go
```diff
@@ -1025,29 +1029,66 @@ func gatherPostgresLogsAndConfigs(ctx context.Context,
...
-			if stderr != "" {
-				str := fmt.Sprintf("\nError returned: %s\n", stderr)
-				buf.Write([]byte(str))
+			// longFileName is namespace/podname:path/to/file
+			longFileName := fmt.Sprintf("%s/%s:%s", namespace, pod.Name, logFile)
+			writeInfo(cmd, fmt.Sprintf("\tSize of %-85s %v", longFileName, ConvertBytes(fileSize)))
+
+			// flags to switch behavior between storing the file in memory versus disk
+			doCat := false
+			doStream := true
```
benjaminjb, line 1045: Are we planning to switch behavior? Looks like we're just going to stream now.

internal/cmd/export.go
```diff
@@ -1025,29 +1029,66 @@ func gatherPostgresLogsAndConfigs(ctx context.Context,
...
-				return err
+			if doStream {
+				err = streamFileFromPod(clientset, config, tw,
+					outputDir, clusterName, namespace, pod.Name, util.ContainerDatabase, logFile)
+
+				if err != nil {
+					writeInfo(cmd, fmt.Sprintf("\tError streaming file %s: %v", logFile, err))
+					// "failed to remove" errors are not remedied with kubectl cp
+					if !strings.Contains(err.Error(), "failed to remove") {
+						writeInfo(cmd, fmt.Sprintf("\tCollect manually with kubectl cp -c %s %s/%s:%s %s",
```
benjaminjb, line 1081: Did we ever talk about the possibility of having the pgo CLI call `kubectl cp` itself?

internal/cmd/export.go
philrhurst: I'd like to remove all the cat behavior; but I wanted to get another set of eyes since it was a pretty big change in behavior.

I will refactor do remove all the 'doCat' stuff.

internal/cmd/export.go
philrhurst: I've thought about it - calling `kubectl` itself (for `cp` and perhaps `describe`) would be quite nice; but I wasn't sure what behavior we'd have to account for (i.e. `kubectl` version skew, etc)

philrhurst: to be clear, I think calling `kubectl` is worth looking into

internal/cmd/export.go
```diff
@@ -1024,30 +1030,56 @@ func gatherPostgresLogsAndConfigs(ctx context.Context,
...
-			buf.Write([]byte(stdout))
-			if stderr != "" {
-				str := fmt.Sprintf("\nError returned: %s\n", stderr)
-				buf.Write([]byte(str))
+			// fileSpecSrc is namespace/podname:path/to/file
+			// fileSpecDest is the local destination of the file
+			// These are used to help the user grab the file manually when necessary
+			fileSpecSrc := fmt.Sprintf("%s/%s:%s", namespace, pod.Name, logFile)
+			fileSpecDest := filepath.Join(localDirectory, logFile)
+			writeInfo(cmd, fmt.Sprintf("\tSize of %-85s %v", fileSpecSrc, ConvertBytes(fileSize)))
```
benjaminjb, line 1056: Why 85?

internal/cmd/export.go
```diff
@@ -1800,3 +1832,185 @@ func writeDebug(cmd *cobra.Command, s string) {
...
+		Param("stderr", "true").
+		Param("stdout", "true")
+
+	exec, err := remotecommand.NewSPDYExecutor(config, "GET", req.URL())
+	if err != nil {
+		return fmt.Errorf("failed to initialize SPDY executor: %w", err)
+	}
+
+	// remoteFileSize is the file size within the container
+	remoteFileSize, err := getRemoteFileSize(clientset, config, namespace, podName, containerName, remotePath)
```
benjaminjb, line 1879: By the time we call `streamFileFromPod`, haven't we gotten the file size?

internal/cmd/export.go
```diff
@@ -1800,3 +1832,185 @@ func writeDebug(cmd *cobra.Command, s string) {
...
+		Param("container", containerName).
+		Param("command", "cat").
+		Param("command", remotePath).
+		Param("stderr", "true").
+		Param("stdout", "true")
+
+	exec, err := remotecommand.NewSPDYExecutor(config, "GET", req.URL())
+	if err != nil {
+		return fmt.Errorf("failed to initialize SPDY executor: %w", err)
+	}
```
benjaminjb, line 1876: Could we use the `util.NewPodExecutor` to set this command up?

internal/cmd/export.go
philrhurst: the `fileSpecSrc` can be quite long and this aligns it nicely

```
Collecting Postgres logs...
        Size of postgres-operator/hippo-instance1-vp9k-0:pgdata/pg16/log/postgresql-5GB.log           5.1 GB
        Size of postgres-operator/hippo-instance1-vp9k-0:pgdata/pg16/log/postgresql-Tue.log           13.7 KB
        Size of postgres-operator/hippo-instance1-vp9k-0:pgdata/pg16/log/postgresql-1GB.log           1.0 GB
        Size of postgres-operator/hippo-instance1-vp9k-0:pgdata/pg16/log/postgresql-500MB.log         524.3 MB
Collecting pgBackRest logs...
```

internal/cmd/export.go
```diff
@@ -1800,3 +1832,185 @@ func writeDebug(cmd *cobra.Command, s string) {
...
+	if err != nil {
+		return fmt.Errorf("error during file streaming: %w", err)
+	}
+
+	// compare file sizes
+	localFileInfo, err := os.Stat(localPath)
+	if err != nil {
+		return fmt.Errorf("failed to get file info: %w", err)
+	}
+	if remoteFileSize != localFileInfo.Size() {
```
benjaminjb, line 1899: I'm trying to think of a case where the files would be different sizes and that would be ok, and that doesn't seem likely, so 👍

internal/cmd/export.go
philrhurst: yep - `remoteFileSize` could be passed into `streamFileFromPod` to save an extra call

internal/cmd/export.go
```diff
@@ -1800,3 +1832,185 @@ func writeDebug(cmd *cobra.Command, s string) {
...
+	}
+
+	// Write header to the tar
+	err = tw.WriteHeader(header)
+	if err != nil {
+		return fmt.Errorf("failed to write tar header: %w", err)
+	}
+
+	// Stream the file content to the tar
+	_, err = io.Copy(tw, file)
```
benjaminjb, line 1950: Have you tried a setup where the workspace was small and the file big and
* you used the existing client and it failed? and
* you used this new client and it succeeded?

internal/cmd/export.go
```diff
@@ -1800,3 +1832,185 @@ func writeDebug(cmd *cobra.Command, s string) {
...
+		Param("command", "-c").
+		Param("command", "%s").
+		Param("command", filePath).
+		Param("stderr", "true").
+		Param("stdout", "true")
+
+	exec, err := remotecommand.NewSPDYExecutor(config, "GET", req.URL())
+	if err != nil {
+		return 0, fmt.Errorf("could not initialize SPDY executor for stat: %w", err)
+	}
```
benjaminjb, line 1982: Ditto question re: util.NewPodExecutor

internal/cmd/export.go
philrhurst: I've refactored to remove the extra call to `getRemoteFileSize`

internal/cmd/export.go
philrhurst: That was my thought as well. With network issues, even the traditional `kubectl cp` will sometimes only download part of the file and complete successfully (or at least not with an easily identified error). Comparing the file size is one way to ensure accuracy. Hashing the files is probably overkill because these could be large PG logs.

internal/cmd/export.go
philrhurst: I refactored this one to use the `bashCommand` function and executor. This one is easier because it's a simple `stat` command.

internal/cmd/export.go
philrhurst: yes

internal/cmd/export.go
philrhurst: The problem I've had there is the current flow assumes the contents will be held in memory. I couldn't figure out how to use the current flow to pass to `*os.File` - Definitely a limitation on my part, so feedback is welcome

internal/cmd/export.go
benjaminjb: I might throw in a comment there -- hmmm, sometimes I see people throw in actual samples of what the output should look like. I wonder if it's worth doing that here? I'm leaning towards no, but just wanted to raise the idea.

philrhurst: good point - there's a slight description at the beginning of the comment, but I'll expand with a full example

internal/cmd/export.go
benjaminjb: hmmm, `NewPodExecutor` returns a func that takes `stdout, stderr io.Writer`, and I'm not against writing a specific function for this executor, like

```
func (exec Executor) copyFile(source string, destination *os.File) (string, error) {
	var stderr bytes.Buffer
        command := fmt.Sprintf("cat %s", source)
	err := exec(nil, &desination, &stderr, "bash", "-ceu", "--", command)
	return stderr.String(), err
}
```

That's a slight variation on the `catFile` func, the main difference being that it sets the output to the file (I think) and doesn't return a stdout. Not sure that would work, but that's my first thought.

internal/cmd/export.go
```diff
@@ -1800,3 +1832,164 @@ func writeDebug(cmd *cobra.Command, s string) {
...
+	size, err := strconv.ParseInt(strings.TrimSpace(stdout), 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse file size: %w", err)
+	}
+
+	return size, nil
+}
+
+// ConvertBytes converts a byte size (int64) into a human-readable format.
+func ConvertBytes(bytes int64) string {
```
benjaminjb, line 1984: We're not exporting this command for another package to use, so `convertBytes` should be fine. (Also here's a case where I think tests would do some work as documentation.)

internal/cmd/export.go
philrhurst: I added a comment with an example
```
// These are used to help the user grab the file manually when necessary
// e.g. postgres-operator/hippo-instance1-vp9k-0:pgdata/pg16/log/postgresql-Tue.log
fileSpecSrc := fmt.Sprintf("%s/%s:%s", namespace, pod.Name, logFile)
fileSpecDest := filepath.Join(localDirectory, logFile)
writeInfo(cmd, fmt.Sprintf("\tSize of %-85s %v", fileSpecSrc, ConvertBytes(fileSize)))
```

internal/cmd/export.go
philrhurst: one slight change `&destination` to `destination`, but I think this works.

```
func (exec Executor) copyFile(source string, destination *os.File) (string, error) {
	var stderr bytes.Buffer
	command := fmt.Sprintf("cat %s", source)
	err := exec(nil, destination, &stderr, "bash", "-ceu", "--", command)
	return stderr.String(), err
}
```

internal/cmd/export.go
philrhurst: I renamed this one

internal/cmd/export.go
```diff
@@ -1800,3 +1832,151 @@ func writeDebug(cmd *cobra.Command, s string) {
...
+			stdin, stdout, stderr, command...)
+	}
+
+	_, err = Executor(exec).copyFile(remotePath, outFile)
+	if err != nil {
+		return fmt.Errorf("error during file streaming: %w", err)
+	}
+
+	// compare file sizes
+	localFileInfo, err := os.Stat(localPath)
```
benjaminjb, line 1877: Here we use `os.Stat(FILE)` and on line 1912 we use `file.Stat()` -- they're functionally the same, right? Not a blocker, just curious if we could use the same func twice. (I'm looking at the code for the two functions and they're not identical in the way I sort of thought they might be, though they both end up calling the `fillFileStatFromSys` helper func.)

benjaminjb (approved):
No blockers, just one last question

PR #115, opened 2024-12-05
Improve Logging and Error Handling

philrhurst:
This PR refactors logic to continue gathering data if encountering errors with support export.

This PR also improves how error messages are reported from Patroni and pgBackRest. The error messages are written to the console and the support export log for troubleshooting.

### DIFF
--- a/internal/cmd/export.go
+++ b/internal/cmd/export.go
@@ -1523,7 +1523,7 @@ func gatherPatroniInfo(ctx context.Context,
 			writeInfo(cmd, err.Error())
 			return nil
 		}
-		writeInfo(cmd, strings.TrimSpace(fmt.Sprintf("Error with patronictl list: %s", stderr)))
+		writeInfo(cmd, strings.TrimSpace(fmt.Sprintf("Error with patronictl list: %s: %s", err, strings.TrimSpace(stderr))))
 	}
 
 	buf.Write([]byte(stdout))
@@ -1538,7 +1538,7 @@ func gatherPatroniInfo(ctx context.Context,
 			writeInfo(cmd, err.Error())
 			return nil
 		}
-		writeInfo(cmd, strings.TrimSpace(fmt.Sprintf("Error with patronictl history: %s", stderr)))
+		writeInfo(cmd, strings.TrimSpace(fmt.Sprintf("Error with patronictl history: %s: %s", err, strings.TrimSpace(stderr))))
 	}
 
 	buf.Write([]byte(stdout))
@@ -1601,7 +1601,7 @@ func gatherPgBackRestInfo(ctx context.Context,
 			writeInfo(cmd, err.Error())
 			return nil
 		}
-		writeInfo(cmd, strings.TrimSpace(fmt.Sprintf("Error with pgbackrest info: %s", stderr)))
+		writeInfo(cmd, fmt.Sprintf("Error with pgbackrest info: %s: %s", err, strings.TrimSpace(stderr)))
 	}
 
 	buf.Write([]byte(stdout))
@@ -1616,7 +1616,7 @@ func gatherPgBackRestInfo(ctx context.Context,
 			writeInfo(cmd, err.Error())
 			return nil
 		}
-		writeInfo(cmd, strings.TrimSpace(fmt.Sprintf("Error with pgbackrest check: %s", stderr)))
+		writeInfo(cmd, fmt.Sprintf("Error with pgbackrest check: %s: %s", err, strings.TrimSpace(stderr)))
 	}
 
 	buf.Write([]byte(stdout))
--- a/internal/cmd/export.go
+++ b/internal/cmd/export.go
@@ -1523,7 +1523,7 @@ func gatherPatroniInfo(ctx context.Context,
 			writeInfo(cmd, err.Error())
 			return nil
 		}
-		writeInfo(cmd, strings.TrimSpace(fmt.Sprintf("Error with patronictl list: %s: %s", err, strings.TrimSpace(stderr))))
+		writeInfo(cmd, fmt.Sprintf("Error with patronictl list: %s: %s", err, strings.TrimSpace(stderr)))
 	}
 
 	buf.Write([]byte(stdout))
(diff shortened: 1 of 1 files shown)


philrhurst:
The linter is failing for a part of the code not touched by this PR. I haven't seen this failure before, so I'm not sure why this check is failing now when it didn't fail before.

```
Error: internal/config.go:25:6: the methods of "PatchConfig" use pointer receiver and non-pointer receiver. (recvcheck)
```

internal/cmd/export.go
```diff
@@ -387,88 +387,111 @@ Collecting PGO CLI logs...
...
+		if err != nil {
+			writeInfo(cmd, fmt.Sprintf("Error gathering Kubernetes server version: %s", err))
 		}
 
-		if err == nil {
-			err = gatherNodes(ctx, clientset, clusterName, tw, cmd)
+		// Gather list of nodes
+		err = gatherNodes(ctx, clientset, clusterName, tw, cmd)
+		if err != nil {
+			writeInfo(cmd, fmt.Sprintf("Error gathering list of cluster nodes: %s", err))
```
tjmoore4, line 415: :pick: Might be worth clarifying that this refers to Kubernetes clusters rather than Postgres clusters.

tjmoore4 (approved):

internal/cmd/export.go
```diff
@@ -526,13 +560,8 @@ Collecting PGO CLI logs...
...
 		// Print final message
-		if err == nil {
-			info, err := os.Stat(outputDir + "/" + outputFile)
-
-			if err == nil {
-				fmt.Print(exportSizeReport(float64(info.Size())))
-			}
-		}
+		info, err := os.Stat(outputDir + "/" + outputFile)
+		fmt.Print(exportSizeReport(float64(info.Size())))
```
benjaminjb, line 564: ?

internal/cmd/export.go
```diff
@@ -1493,7 +1523,7 @@ func gatherPatroniInfo(ctx context.Context,
 			writeInfo(cmd, err.Error())
 			return nil
 		}
-		return err
+		writeInfo(cmd, strings.TrimSpace(fmt.Sprintf("Error with patronictl list: %s", stderr)))
```
benjaminjb, line 1526: Are we sure we want to print stderr and drop err?

internal/cmd/export.go
philrhurst: good point - changed to `Error gathering list of Kubernetes nodes`

internal/cmd/export.go
philrhurst: That's what causes this to be printed as the final message. It's been in the product for a while. A bit of a visual reminder for the user to know if the Support Export is too-big-to-email. There's actually an `if` statement with the `exportSizeReport` function to specifically tell the user what to do if the size is larger than 25MB.

```
Collecting PGO CLI logs...
┌────────────────────────────────────────────────────────────────
| Archive file size: 0.11 MiB
| Email the support export archive to support@crunchydata.com
| or attach as a email reply to your existing Support Ticket
└────────────────────────────────────────────────────────────────
```

internal/cmd/export.go
philrhurst: I think so - the underlying issue is that when/if an error happens, the Support Export stops collecting any more data.

So in this specific case, if there is an issue running `patronictl list`, no further information is gathered. That is not great when the subsequent data points are necessary. I think we should print the patroni error message (patroni could report helpful info on _why_ the command failed) and continue on. In mind, it's not appropriate to `return err` because the function can still recover at this error. But I'm open to discussion.

benjaminjb:
> > The linter is failing for a part of the code not touched by this PR. I haven't seen this failure before, so I'm not sure why this check is failing now when it didn't fail before.
> > ```
> > Error: internal/config.go:25:6: the methods of "PatchConfig" use pointer receiver and non-pointer receiver. (recvcheck)
> > ```
> 
> We pull in the [latest](https://github.com/CrunchyData/postgres-operator-client/blob/main/.github/workflows/lint.yaml#L21) version of the golang linter when we do our checks and the and version 1.62 added [recvcheck](https://github.com/golangci/golangci-lint/blob/master/CHANGELOG.md#v1620). The error is because `AddFlags` uses a pointer receiver and the other methods use non-pointer receivers. `AddFlags` doesn't seem to be used at this time, so the easiest option would be to use a non-pointer receiver there, but I'll double check with the team whether or not that's what we want to do. Feel free to disregard for now.

A) Agree with TJ here re: why this is being picked up now (e.g., floating linter version means we run most recent in GH actions; these methods mix pointer and value receivers)
B) Agree with TJ here re: feel free to disregard for now.
C) Think I disagree with TJ re: fix -- rather than go all non-pointer receivers, I think we should go all pointer receivers.

It's funny that this just came up now because recently Chris went through pgo for a similar issue, so I've been reviewing [this section](https://go.dev/wiki/CodeReviewComments#receiver-type) a bunch.

The principles in that doc are (in order of importance)
- be consistent
- use pointers when in doubt (especially if you're going to mutate the receiver)
- in other cases, especially when the receiver is small, you could use a value receiver

In `CreateOptions` and `PatchOptions`, we're not mutating the receiver -- we're using the receiver to mutate the input, i.e., to add a field to the input struct. `AddFlags` does mutate the receiver, though (we set the string which is a field on the receiver struct).

internal/cmd/export.go
benjaminjb: Oh, yeah, I see the need to not quit at every error. I just wonder if `stderr` and `err` might both have different and useful info -- should we print both? I don't remember (off the top of my head) what sort of errs and stderrs we get from these execs.

philrhurst: In my investigation, `stderr` tends to have more app-specific info. My case was failing `pgbackrest` commands (which is what prompted this original issue). `err` tended to only report the error number, so it required detective work to determine what subsystem was even associated with the error. 

I've got some ideas to combine the output from `err` and `stderr` in a meaningful way.

philrhurst: Here's how things look when I combine the patroni and pgbackrest sections with something like `writeInfo(cmd, strings.TrimSpace(fmt.Sprintf("Error with patronictl list: %s: %s", err, strings.TrimSpace(stderr))))`

`stderr` needs to have its whitespace trimmed and putting `err` first makes things look better when reading the output.

```
Collecting Patroni info...
Error with patronictl list: command terminated with exit code 2: Usage: patronictl [OPTIONS] COMMAND [ARGS]...
Try 'patronictl --help' for help.

Error: No such command 'foo'.
Error with patronictl history: command terminated with exit code 2: Usage: patronictl [OPTIONS] COMMAND [ARGS]...
Try 'patronictl --help' for help.

Error: No such command 'foo'.
Collecting pgBackRest info...
Error with pgbackrest info: command terminated with exit code 48: ERROR: [048]: invalid command 'foo'
Error with pgbackrest check: command terminated with exit code 48: ERROR: [048]: invalid command 'foo'
Collecting processes...
```

I can't do much to change how the subsystems report their errors, but the exact output is presented to the user for further investigation.

benjaminjb (approved):

PR #117, opened 2024-12-16
Update the support export tool to gather Patroni logs

tjmoore4:
Ensures that the on volume Patroni log file is exported, if that file exists. If the PostgresCluster is not configured to create this file, note in the debug logs that this is acceptable for some configurations.

Issue: PGO-1701


internal/cmd/export.go
```diff
@@ -1308,6 +1314,113 @@ func gatherDbBackrestLogs(ctx context.Context,
...
+			if err != nil {
+				writeDebug(cmd, fmt.Sprintf("%s\n", err.Error()))
+			}
+			if stderr != "" {
+				writeDebug(cmd, stderr)
+			}
+
+			if strings.Contains(stderr, "No such file or directory") {
+				writeDebug(cmd, "Cannot find any Patroni log files. This is acceptable in some configurations.\n")
+			}
```
benjaminjb, line 1388: If the stderr is `No such file or directory`, we'll print two lines for this one err, right? Is that what we want?

benjaminjb (commented):
Do you think it's worth adding to the support export kuttl test?

internal/cmd/export.go
tjmoore4: I think so. It follows are existing pattern and only shows up in the `DEBUG` logs. For reference, here's what you would see in `cli.log`:
```
2024-12-18 15:15:31.620 -0500 EST - DEBUG - Error getting Patroni logs
2024-12-18 15:15:31.620 -0500 EST - DEBUG - command terminated with exit code 2
2024-12-18 15:15:31.620 -0500 EST - DEBUG - ls: cannot access 'pgdata/patroni/log/*': No such file or directory
2024-12-18 15:15:31.620 -0500 EST - DEBUG - Cannot find any Patroni log files. This is acceptable in some configurations.
```

tjmoore4:
> Do you think it's worth adding to the support export kuttl test?

I considered it but it doesn't look like the goal of that test is to test every item we collect, especially when there's little variance in what's being done. Perhaps the approach should be updated?

internal/cmd/export.go
benjaminjb: Hmmm, if that's the pattern we've got, I'm fine with it (with the plan to come back and review our logging patterns).

internal/cmd/export.go
```diff
@@ -1308,6 +1314,113 @@ func gatherDbBackrestLogs(ctx context.Context,
...
+// Pod logs instead.
+func gatherPatroniLogs(ctx context.Context,
+	clientset *kubernetes.Clientset,
+	config *rest.Config,
+	namespace string,
+	clusterName string,
+	tw *tar.Writer,
+	cmd *cobra.Command,
+) error {
+	writeInfo(cmd, "Collecting Patroni logs...")
```
benjaminjb, line 1328: Do we want to throw this line into the help text (around line 262) and then regen the docs?

benjaminjb (approved):
Curious if @philrhurst has any feedback on this.

No blockers for me, just a question about our help text/documentation.

internal/cmd/export.go
tjmoore4: Yep, I'll make that change. Thanks!

tjmoore4: Updated, plus I added a couple other missing log references.

internal/cmd/export.go
philrhurst: In the other cases, the check at `strings.Contains(stderr, "No such file or directory")` was necessary because the _lack_ of logs was not necessarily an erroneous event. That extra information confirms there are no files, not that we encountered an error getting them.

(If I remember correctly)

internal/cmd/export.go
```diff
@@ -1308,6 +1317,113 @@ func gatherDbBackrestLogs(ctx context.Context,
...
+			}
+			continue
+		}
+
+		logFiles := strings.Split(strings.TrimSpace(stdout), "\n")
+		for _, logFile := range logFiles {
+			writeDebug(cmd, fmt.Sprintf("LOG FILE: %s\n", logFile))
+			var buf bytes.Buffer
+
+			stdout, stderr, err := Executor(exec).catFile(logFile)
```
philrhurst, line 1400: I'm fine with this call with `catFile`, but I am concerned that as the Patroni logs grow we will encounter issues similar to the Postgres logs: the entire file cannot be held in memory. We stream the PG logs to local disk to get around that issue. We've got good error handling here, so I don't think we should refactor for that; but it's something to keep in mind.

tjmoore4: For the Patroni logs, there is a defined storage limit, but I agree this could still be a problem for larger file sizes. I'm wondering if the right answer is to do some refactoring and combine the various functions that pull logs from the instance Pods so they all benefit from the streaming implementation?

tjmoore4: Since this isn't a blocker, I'm going to merge this PR now but I'll (hopefully) have a refactor PR up soon for consideration.

PR #127, opened 2025-05-20
Adjust support for OTel and non-OTel

benjaminjb:
With OTel, we've introduced some changes, most notably to the location of the Postgres logs; but also with the introduction of a `receiver` dir in several locations which only the collector container has access to.

This PR addresses these issues by using the cluster to determine the location of the logs, and by excluding checking the `receiver` dir in several execs through the use of a '*.*' wildcard.

Issues: [PGO-2358]


philrhurst (approved):
LGTM

PR #138, opened 2025-08-27
Migrating PGO CLI kuttl tests to chainsaw

philrhurst:
These are the first round of kuttl tests migrated to chainsaw. The structure is a single test for each kubectl-pgo command

- backup
- create
- delete
- show
- version


tjmoore4 (commented):
Looks good overall. A few questions.

testing/chainsaw/e2e/backup/chainsaw-test.yaml
```diff
@@ -0,0 +1,705 @@
...
+          duration: 10s
+
+    - name: 'Confirm cluster is created'
+      use:
+        template: '../templates/confirm-created.yaml'
+
+    - name: Sleep for 10s
+      try:
+      - sleep:
+          duration: 10s
```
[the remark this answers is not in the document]
benjaminjb: Can we set timeouts on different steps to give longer timeouts for the confirm steps or whatever that is taking so long.

testing/chainsaw/e2e/create/chainsaw-test.yaml
```diff
@@ -0,0 +1,146 @@
...
+    - name: Sleep for 30s
+      try:
+      - sleep:
+          duration: 30s
+
+    - name: 'Confirm cluster is created'
+      use:
+        template: '../templates/confirm-created.yaml'
+
+    - name: Sleep for 30s
```
benjaminjb, line 48: If we need these sleep steps, I think it would be fine to name them all `Sleep` or something without the time -- it's easy enough to look at the duration to see how long and it avoids drift when we change those durations.

More interesting to me would be why we chose those durations.

testing/chainsaw/e2e/start-stop/chainsaw-test.yaml
```diff
@@ -0,0 +1,255 @@
...
+              value: ($cluster)
+          entrypoint: "sh"
+          args:
+          - "-c"
+          - "echo 'n' | kubectl-pgo stop $CLUSTER --force-conflicts --namespace=$NAMESPACE"
+          timeout: 10s
+
+    - name: 'Confirm cluster is created'
+      use:
+        template: '../templates/confirm-created.yaml'
```
benjaminjb, line 54: So the idea is that we can use the same template here because the command did nothing?

testing/chainsaw/e2e/start-stop/chainsaw-test.yaml
```diff
@@ -0,0 +1,255 @@
...
+
+    - name: "Sleep 30s"
+      try:
+        - sleep:
+            duration: 30s
+
+    - name: run 'stop cluster' with confirm 'y' and force-conflicts
```
benjaminjb, line 152: if the `start` test runs the stop command, I think we can collapse this into one test, yeah?

testing/chainsaw/policies/v1.15.1/00-rbac.yaml
```diff
@@ -0,0 +1,32 @@
...
+    rbac.kyverno.io/aggregate-to-background-controller: "true"
+rules:
+- apiGroups:
+  - ''
+  resources:
+  - secrets
+  verbs:
+  - create
+  - update
+  - delete
```
benjaminjb, line 30: Why delete?

philrhurst: I was having some issues getting it to work. This Github comment suggested delete.

- https://github.com/kyverno/kyverno/issues/11467#issuecomment-2480926228

testing/chainsaw/e2e/backup/chainsaw-test.yaml
```diff
@@ -0,0 +1,705 @@
...
+
+    - name: Verify COMMAND_OPTS matches "--stanza=db --repo=1"
+      bindings:
+        - name: selector
+          value: (replace_all(
```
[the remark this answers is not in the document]
philrhurst: yes - it was a bit of over-engineering. The backupSelector is constant for the entire test; so we can just define it fully.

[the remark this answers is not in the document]
philrhurst: I've refactored to simplify the select injection. No more nesting!

testing/chainsaw/e2e/backup/chainsaw-test.yaml
```diff
@@ -0,0 +1,705 @@
...
+          timeout: 300s
+          for:
+            jsonPath:
+              path: '{.status.phase}'
+              value: 'Succeeded'
+---
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: check-backup-command-no-flags
```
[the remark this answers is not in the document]
philrhurst: This is the conversion of `08--backup-with-just-trigger.yaml` which is triggering a backup from the PGO CLI with no options and checking the annotations

testing/chainsaw/e2e/create/chainsaw-test.yaml
philrhurst: Most of the Sleep 30s started at Sleep 10. I bumped them up because running all the test simultaneously was causing issues. I left the descriptive "Sleep for 30s" as I was debugging the tests to know how much time I should wait before the next step.

testing/chainsaw/e2e/restore/chainsaw-test.yaml
```diff
@@ -0,0 +1,713 @@
...
+              name: ($cluster)
+            spec:
+              backups:
+                pgbackrest:
+                  restore: {}
+---
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: check-restore-confirm-yes
```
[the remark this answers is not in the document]
philrhurst: Yes it is, but each "test" would then be a separate "step". It becomes an issue of clarity and preference. I.e. do we have a single Restore Test that does multiple things. Or multiple Restore Tests, each doing one single thing.

One thing I like about Chainsaw is that each test is namespace-scoped. So it's easy to separate out tests into discrete units.

testing/chainsaw/e2e/restore/chainsaw-test.yaml
```diff
@@ -0,0 +1,713 @@
...
+          entrypoint: "sh"
+          args:
+          - "-c"
+          - "echo yes | kubectl-pgo restore $CLUSTER -n $NAMESPACE --repoName repo1"
+          timeout: 10s
+          check:
+            (contains($stderr, 'Apply failed')): true
+            (contains($stderr, '2 conflicts')): true
+
+    - name: run 'restore' with confirm 'yes' and --force-conflicts
```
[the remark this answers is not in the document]
philrhurst: yes, I will look into this. I think I cribbed more of the original kuttl test on this particular one than the others.

testing/chainsaw/e2e/values.yaml
```diff
@@ -0,0 +1,5 @@
+versions:
+  postgres: '17'
+
+images:
+  psql: 'registry.developers.crunchydata.com/crunchydata/crunchy-postgres:ubi9-17.5-2520'
```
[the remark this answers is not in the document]
philrhurst: It probably should be - I was basing this on an example we were using elsewhere.

testing/chainsaw/e2e/version/chainsaw-test.yaml
```diff
@@ -0,0 +1,50 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  creationTimestamp: null
+  name: version
+spec:
+  steps:
+  - name: step-00
```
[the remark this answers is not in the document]
philrhurst: sure :)

tjmoore4 (approved):
Updates look good to me. Not sure if @benjaminjb is still tracking any blockers, but I think this is good to merge.

testing/chainsaw/e2e/support/chainsaw-test.yaml
```diff
@@ -0,0 +1,932 @@
...
+kind: Test
+metadata:
+  name: kuttl-support-cluster
+spec:
+  bindings:
+    - name: cluster
+      value: kuttl-support-cluster
+
+    - name: postgresVersion
+      value: 16
```
benjaminjb, line 11: Don't we have a central source for pg version?

testing/chainsaw/e2e/support/chainsaw-test.yaml
```diff
@@ -0,0 +1,932 @@
...
+    - name: 'Create Monitoring Cluster'
+      try:
+      - apply:
+          resource:
+            apiVersion: postgres-operator.crunchydata.com/v1beta1
+            kind: PostgresCluster
+            metadata:
+              name: kuttl-support-monitoring-cluster
+            spec:
+              postgresVersion: 16
```
benjaminjb, line 419: and if we have a central location for the pg version (or even a local version), better to use it

testing/chainsaw/e2e/support/chainsaw-test.yaml
```diff
@@ -0,0 +1,932 @@
...
+                              storage: 1Gi
+              instances:
+                - dataVolumeClaimSpec:
+                    accessModes:
+                    - "ReadWriteOnce"
+                    resources:
+                      requests:
+                        storage: 1Gi
+                  replicas: 1
+              postgresVersion: 16
```
benjaminjb, line 731: ditto

testing/chainsaw/e2e/support/chainsaw-test.yaml
philrhurst: Yes we do; but the bash for this particular test has already hard-coded 16 in it. I didn't want to refactor to interrupt migrating the tests.

benjaminjb (approved):
Really no blockers, but re-reading these tests -- many of which I probably had a hand in -- makes me want to revise these tests. Please make a ticket so we don't forget to!

(What I really want is to shift tests left, e.g., run as much as we can as unit tests) and also combine a few of these tests if we can to run them faster. I know I have created a few tickets re: separating out the cluster work from some CLI work, so it won't just be test changes I'm envisioning in the future.)

